PR #12, opened 2016-03-31
Add GRU

ereday:


denizyuret:
Small recommendations:
- It doesn't look like you are using the fbias option, you can remove it.
- For negating etc. consider using the axpb operator which applies a*x^p+b elementwise.  Documented at http://knet.readthedocs.org/en/latest/intro.html#some-useful-tables.  By setting the a, p, b keyword options you can do quite a bit with this function and it is efficient.

ereday:
Make changes in accordance with the recommendations. 
1.  Remove unused `fbias` parameter from the parameter list of GRU 
2.  Use `axpb` operator instead of  separate element-wise multiplication and addition
3.  Use` mul `and `add` that are alternative notations   of  `.*` and `.+` respectively.

PR #13, opened 2016-04-06
Implementation of transpose operation 

kirnap:


denizyuret:
Merging this to the transpose-0.7 branch.

PR #14, opened 2016-04-28
Fix 1D pooling

quaertym:


denizyuret:
Merged this in both master and release-0.7.

PR #17, opened 2016-06-10
Fix deprecation for julia v0.5

rejuvyesh:
Explicit `Dict(a=>b)` instead of `[a=>b]`.


denizyuret:
Thanks for the pull request.  This seems to fail on Julia 0.4.5, can you fix it so it works in both v4 and v5?

rejuvyesh:
It seems `Compat` does not have an equivalent yet. I'll see what can be done.

denizyuret:
There are further problems with Knet release-0.7 supporting Julia5.  I marked this branch as compatible with Julia4 only.  The latest version (Knet 0.8) should be compatible with Julia4, 5 and 6.

PR #21, opened 2016-07-11
empty error message

melikesofta:
empty error message filled for when sback is called too many times on a stack


denizyuret:
merged with the branch release-0.7.

PR #28, opened 2016-10-20
Sgd based optimization methods

ozanarkancan:


denizyuret:
A couple of minor comments:
- are the init_x functions necessary?  we should just use the constructors for your types.  just rename init_adam => Adam etc.  Less symbols to export.
- I like it when gc() prints '.', why did you delete it :) ?

denizyuret:
convert(typeof(w), zeros(size(w))) allocates unnecessary memory, use zeros(w) or fill!(similar(w),0).

Issue #39, opened 2016-11-12
KnetArray transpose and permutedims missing.


denizyuret:

denizyuret:
Is permutedims for 4D and 5D practical?  Necessary?

denizyuret:
I am leaving this issue open for the general N-D implementation.

denizyuret:
@ilkerkesen do we actually have a general N-D implementation (beyond 5-D) that works?  If so, let's close this.

ilkerkesen:
Back then I implemented a general permutedims operation, but couldn't port
it to Knet. Yes, we have, but it needs to be ported.

On Dec 24, 2017 6:55 PM, "denizyuret" <notifications@github.com> wrote:

@ilkerkesen <https://github.com/ilkerkesen> do we actually have a general
N-D implementation (beyond 5-D) that works? If so, let's close this.

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/denizyuret/Knet.jl/issues/39#issuecomment-353791444>,
or mute the thread

ilkerkesen:
Caffe2 has a [multi-dimensional transpose operation](https://caffe2.ai/docs/operators-catalogue#transpose) in the way we desire. It takes advantage of cudnn's `cudnnTransformTensor` kernel. I'm on it.

denizyuret:
Solved using CuArrays under permutedims branch, please test.

PR #91, opened 2017-02-27
N-D permutedims

ekyurdakul:
Completed #39. 2D, 3D, 4D, 5D permutedims implemented.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,6 +1,7 @@
 julia 0.4
 AutoGrad 0.0.6
 Compat 0.16.0   # for @views
+Combinatorics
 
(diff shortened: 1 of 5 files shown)


ilkerkesen:
@denizyuret is it okay to add Combinatorics package as a new dependency? It would be good for our general array tests. For instance, I'd like to test all possible reductions in a 5-D array ((1,2),(1,3),...,(4,5),(1,2,3),...,(3,4,5),(1,2,3,4)...).

denizyuret:
The test directory has its own REQUIRE you can certainly add it there.  If
this is the only function you need, I would just write the function (or
copy it from Combinatorics) to eliminate the dependency.

PR #102, opened 2017-03-23
make vgg/resnet examples working on cpu

ilkerkesen:
I just did it like how it is done in lenet example.

### DIFF
--- a/examples/resnet.jl
+++ b/examples/resnet.jl
@@ -136,7 +136,7 @@ function batchnorm(w, x, ms; mode=1, epsilon=1e-5)
         d = ndims(x) == 4 ? (1,2,4) : (2,)
         s = reduce((a,b)->a*size(x,b), d)
         mu = sum(x,d) / s
-        sigma = sqrt(epsilon + (sum(x.-mu,d).^2) / s)
+        sigma = sqrt(epsilon + (sum((x.-mu).^2, d)) / s)
     elseif mode == 1
         mu = shift!(ms)
         sigma = shift!(ms)
--- a/examples/resnet.jl
+++ b/examples/resnet.jl
@@ -136,7 +136,9 @@ function batchnorm(w, x, ms; mode=1, epsilon=1e-5)
         d = ndims(x) == 4 ? (1,2,4) : (2,)
         s = prod(size(x,d...))
         mu = sum(x,d) / s
-        sigma = sqrt(epsilon + (sum((x.-mu).^2, d)) / s)
+        x0 = x .- mu
+        x1 = x0 .* x0
+        sigma = sqrt(epsilon + (sum(x1, d)) / s)
     elseif mode == 1
         mu = shift!(ms)
         sigma = shift!(ms)
--- a/examples/resnet.jl
+++ b/examples/resnet.jl
@@ -134,7 +134,7 @@ function batchnorm(w, x, ms; mode=1, epsilon=1e-5)
     mu, sigma = nothing, nothing
     if mode == 0
         d = ndims(x) == 4 ? (1,2,4) : (2,)
-        s = reduce((a,b)->a*size(x,b), d)
+        s = prod(size(x)[[d...]])
         mu = sum(x,d) / s
         sigma = sqrt(epsilon + (sum((x.-mu).^2, d)) / s)
     elseif mode == 1
--- a/examples/resnet.jl
+++ b/examples/resnet.jl
@@ -134,7 +134,7 @@ function batchnorm(w, x, ms; mode=1, epsilon=1e-5)
     mu, sigma = nothing, nothing
     if mode == 0
         d = ndims(x) == 4 ? (1,2,4) : (2,)
-        s = prod(size(x)[[d...]])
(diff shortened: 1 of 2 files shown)


CarloLucibello:
This is something not touched by this PR, but since you are at this...
In the batchnorm method is
```julia
sigma = sqrt(epsilon + (sum(x.-mu,d).^2) / s)
```
Shouldn't rather be 
```julia
sigma = sqrt(epsilon + (sum((x.-mu).^2, d)) / s)
```
? note: with some thought the last line could be made more efficient

ilkerkesen:
You're right Carlo. Thanks, I've just fixed it.

CarloLucibello:
Also the quantity s is miscalculated I fear (it is for sure in 2d)

Since batchnorm is a useful tool could it be added, tested and exported by
Knet?

Il sab 25 mar 2017, 17:17 İlker Kesen <notifications@github.com> ha scritto:

ilkerkesen:
I followed [this implementation](https://github.com/cthorey/CS231/blob/master/assignment2/cs231n/layers.py#L571) and also Bengio's DL book. I don't see any difference right now. If you think this is the wrong day of doing this, then how should we handle batch normalization?

No, we didn't tested it yet in the training phase, since currently we don't have any models trained by taking advantage of batch normalization layers. These examples demonstrate testing phase.

CarloLucibello:
well, I'm not very experienced with convolutional layers, not with other frameworks, but for 2d this is clearly wrong
```julia
julia> x=rand(7,10);

julia> d = ndims(x) == 4 ? (1,2,4) : (2,)
(2,)

julia> s = reduce((a,b)->a*size(x,b), d)
2
```
A correct implementation could be
```
julia> x=rand(7,10);

julia> d = ndims(x) == 4 ? (1,2,4) : (2,)
(2,)

julia> s = prod(size(x)[[d...]])           
10

julia> x=rand(20,30,40,10);

julia> d = ndims(x) == 4 ? (1,2,4) : (2,)
(1,2,4)

julia> s = prod(size(x)[[d...]])
6000
```

ilkerkesen:
I see, I did misunderstand you. I'll fix that silly mistake of mine, too. Thanks a lot.

denizyuret:
Questions on the resnet implementation before I merge this:

L81. has a bias added after conv.  The paper and the book says bias unnecessary if you are using BN. Is that not the case with convolutional layers?  Convolutional bias is channel-wise, isn't BN bias also channel-wise?  In L153, the only other place BN is used, there is no bias.

L191. I assume "moments" are the training averages of mean/std to use during test (mode=1). What is "bn" for? Are these the gamma/beta parameters you apply after the normalization, i.e. the w arg to batchnorm?

L133. I don't understand shifting mu,sigma in and out of the ms parameter.  If mode==0 you compute mu,sigma, then push them onto ms (why?).  If mode==1 you first shift them out and then push them back in (why?).

L137. prod(size(x,d...)) would be simpler, please confirm that is what you need.

Finally when I try to run these with latest updates I get a "channelview not defined" error.  This may be related to the latest breaking changes in Images.jl -- the NEWS.md has some note about channelview but I could not install the ImageCore pkg mentioned.  Maybe we can find a workaround that does not use channelview?

ilkerkesen:
L81: I followed [the original implementation](https://github.com/KaimingHe/deep-residual-networks) of ResNet shared by the original author. Yes, the only bias parameter applied after convolution is that one in that architecture.

L191: Yes, those parameters are gamma/beta parameters.

L133: Since we perform minibatch training, we take running means of mean/sigma values in the training. So, mode=0 means training, I push them an array, then handle calculating running means of mean/sigma values (also known as moments) in the outside of the prediction function. However, it is not included in this code snippet because we're just handling the test phase. The phase/mode does not matter, I always shift and push them (circulation), because in this way I don't need to handle indexes. Otherwise, things become complicated in the test phase. 

L137: Yes, they are same and yours is much more simpler.

For the channelview, let me take a look at it. I guess they made non back compatible changes recently.

denizyuret:
L133: just to make sure I understand: training uses the actual mean std of
the minibatch, you only need running averages for test phase, right? And in
this demo we never run training so that part never gets used anyway. I
still don't understand the push! and shift! -- why not just keep around a
running count and sum? Do you have a finite sized buffer where you hold the
last N values? If that is the case an alternative would be to keep a moving
average, i.e. const*newval+(1-const)*oldmean.

ilkerkesen:
Yes, I need those mu/sigma values just in the test phase. Yes, you can calculate moving average, but I did think that I can do it outside of the network propagation. At that time I just imagined like that I have previous moving means and std deviations in one array and I collect the mean and stddev of the current minibatch and then update moving means and std deviations.

I use ms array as a circular buffer in the test phase. If I retrieve mu and sigma values by getindex method, then I need to keep track indexes or pass subset of that array to the batch normalization layers (ms[1:2] for the first batch normalization layer). However, in this way I don't have to keep track of indexes or pass the subset corresponds to the current batch normalization layer as an input. After the propagation finished, I will still have the moments in the desired order and the phase doesn't matter.

PR #107, opened 2017-04-04
fix vector reduction correctness

ilkerkesen:
I've just fixed the vector reduction kernel. The problem occurs when we exceed the maximum number of blocks allowed by the architecture. In one block, we need to iterate over many array regions. I just simply added a for loop for that purpose. I've fixed the correctness issue, but this changes made reduction operations a little bit slower as expected. I think optimizing numblocks/numthreads could give the desired improvements in vector reduction.

### DIFF
--- a/src/cuda21.jl
+++ b/src/cuda21.jl
@@ -25,48 +25,49 @@ __global__ void _$(F)_21(int nx, $T *x, int sy, int ny, $T *y) {
   int b = blockIdx.x;
   $T ai, xi;
 
-  // sum the elements assigned to this thread
-  ai = $v0;
-  if (sy == 1) {
-     int istep = $THR*ny;
-     for (int i=b+t*ny; i<nx; i+=istep) {
-        xi=x[i]; xi=$f1; ai=$op;
-     }
-  } else {
-    int jstep = sy*ny;
-    for (int j=0; j<nx; j+=jstep) {
-      int i0 = j+b*sy;
-      int i1 = i0+sy;
-      for (int i=i0+t; i<i1; i+=$THR) {
-        xi=x[i]; xi=$f1; ai=$op;
+  for (int bi = b; bi < ny; bi += blockDim.x) {
+    // sum the elements assigned to this thread
+    ai = $v0;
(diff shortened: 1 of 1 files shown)


denizyuret:
I did some benchmarking with Ilker's fixed reduction kernels.  You can find the code I used in Knet/prof/reduction.jl.  The three tables below show the times (in ns) for arrays of different row and column sizes:

```
sum(a,(1,2))
        1       10      100     1000
1       35243   33322   33538   33617
10      32848   32999   32918   33449
100     32868   33065   33260   33987
1000    31866   31867   34286   71090
sum(a,(1,))
        1       10      100     1000
1       30391   5365    5475    10180
10      30439   5429    5498    11718
100     30205   5469    5421    12082
1000    30458   5849    6460    39497
sum(a,(2,))
        1       10      100     1000
1       30255   30106   30227   30281
10      5419    5499    5420    7112
100     5330    5494    6154    24916
1000    10174   11216   26602   171226
```

Some observations:
* The scalar sum is almost constant time, except for the 1000x1000 array.
* Single column case is very expensive for sum(a,1).
* Single row case is very expensive for sum(a,2).
* The (1000,1000), (100,1000), (1000,100) are abnormally expensive for sum(a,2)

Here are the results with the new kernels:
```
sum(a,(1,2))
        1       10      100     1000
1       32113   32451   32428   31985
10      32059   32223   32159   32388
100     32112   32100   33014   33502
1000    31181   31299   33297   70593
sum(a,(1,))
        1       10      100     1000
1       29654   5070    5071    13076
10      29727   5064    4996    15647
100     30073   4917    5044    16353
1000    29756   5937    6582    45292
sum(a,(2,))
        1       10      100     1000
1       29872   29895   29776   29741
10      5027    5145    5027    7281
100     4961    5054    6197    25029
1000    13093   14504   30535   177023
```

A couple of observations:
* The 1000 column cases are significantly more expensive than before for sum(a,1).
* The 1000 row cases are significantly more expensive than before for sum(a,2).

Enis, can you explain these observations?  Suggest some easy fixes?

PR #111, opened 2017-04-14
N-D Reductions

ilkerkesen:
N-D reductions are okay, I added tests but I've got some problems with tests.

1. prod function (and some others) is not able to pass gradcheck test since its value goes to infinity in big arrays. However, in small arrays we don't have any problem. This applies both Array and KnetArray, does not related to CUDA kernels.
2. gradcheck tests fail because this branch does not have general broadcast operations.

Finally, I've fixed another mistake of mine while writing tests. General reduction kernels seem okay to me but we need to decide how to test them. Right now, I test all the possible combinations in all the reduction ops, but some of them will fail always with the current configuration.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -10,7 +10,7 @@ endif
 ifeq ($(NVCC),0)
   CXX := nvcc
   CFLAGS := -O3 --use_fast_math -Wno-deprecated-gpu-targets --compiler-options "$(CFLAGS)"
-  OBJ := cuda1.o cuda01.o cuda11.o cuda12.o cuda20.o cuda21.o $(OBJ)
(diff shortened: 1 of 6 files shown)


denizyuret:
Enis is finishing general reductions, so we can test them together.  Don't worry about gradcheck with large arrays, see what I did in the latest master/test/reduction.jl: I only run Array KnetArray comparisons for those.  As long as we have isapprox(Array,KnetArray) in results and gradcheck with small arrays pass we are good.  You can mark failing tests temporarily with:
https://docs.julialang.org/en/latest/stdlib/test.html#Broken-Tests-1

denizyuret:
So we have 0, 1, and N-1 dimensional reductions here right?  Did you implement the general reduction by iterating the N-1 case?

PR #140, opened 2017-07-01
[RFC] Batch normalization + LeNet with batchnorm example

### LINKED ISSUE #139: add batchnorm
Could the code from resnet example 
https://github.com/denizyuret/Knet.jl/blob/f3c2887c4f0b1cde6e571aa35f097ccc3710ebc0/examples/resnet.jl
for batch normalization be added and exported by Knet?

```julia
# Batch Normalization Layer
# works both for convolutional and fully connected layers
# mode, 0=>train, 1=>test
function batchnorm(w, x, ms; mode=1, epsilon=1e-5)
    mu, sigma = nothing, nothing
    if mode == 0
        d = ndims(x) == 4 ? (1,2,4) : (2,)
        s = prod(size(x,d...))
        mu = sum(x,d) / s
        x0 = x .- mu
        x1 = x0 .* x0
        sigma = sqrt(epsilon + (sum(x1, d)) / s)
    elseif mode == 1
        mu = shift!(ms)
        sigma = shift!(ms)
    end

    # we need getval in backpropagation
    push!(ms, AutoGrad.getval(mu), AutoGrad.getval(sigma))
    xhat = (x.-mu) ./ sigma
    return w[1] .* xhat .+ w[2]
end
```
I can eventually file a PR.

It is not clear to me how the storage of the momenta, in `ms`, could be handled in a robust way though

CarloLucibello:
I implemented both the version wich keeps the moment of each batch in the training set and the one with exponential decay average (now commented out), following some indications in the discussion in #139 

In my experiments on cpu with the LeNet example the extra operations needed by batchnorm an increase of 
+160% for decaying average
+130% for "keep each batch moments"
in computational time for each epoch with respect to `lenet.jl` example. 
With decaying average we spare on memory resources, which can have some impact on really large training set with small batchsizes.  Which one should I keep?

@denizyuret  @ilkerkesen @AStupidBear @mambuDL  any comments?

Exports the `batchnorm` function and the `BatchMoments` type. fixes #139

### DIFF
--- a/examples/lenet.jl
+++ b/examples/lenet.jl
@@ -165,5 +165,5 @@ end # module
 #
 # lenet.jl (c) Deniz Yuret, 2016. The LeNet model on the MNIST handwritten digit recognition problem from http://yann.lecun.com/exdb/mnist.
 # opts=(:seed,-1)(:batchsize,100)(:epochs,3)(:lr,0.1)(:gcheck,0)(:fast,true)
-# ..................  
+# ..................
 # 9.319163 seconds (5.84 M allocations: 277.927 MB, 7.37% gc time)
--- a/examples/lenet_batchnorm.jl
+++ b/examples/lenet_batchnorm.jl
@@ -110,8 +110,8 @@ end
 """
 function main(;
         seed = -1,
-        batchsize = 100,
-        optimizer = Momentum(lr=0.01),
+        batchsize = 128,
+        optimizer = Momentum(lr=0.1),
         epochs = 100,
         infotime = 1,  # report every `infotime` epochs
         atype = gpu() >= 0 ? KnetArray{Float32} : Array{Float32},
--- a/examples/lenet_batchnorm.jl
+++ b/examples/lenet_batchnorm.jl
@@ -1,0 +1,145 @@
+"""
+This example learns to classify hand-written digits from the
+[MNIST](http://yann.lecun.com/exdb/mnist) dataset.  There are 60000
+training and 10000 test examples. Each input x consists of 784 pixels
+representing a 28x28 image.  The pixel values are normalized to
+[0,1]. Each output y is converted to a ten-dimensional one-hot vector
+(a vector that has a single non-zero component) indicating the correct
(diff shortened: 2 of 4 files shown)


jgbos:
When it comes to modules I don't know if you'd be interested in what I have been playing with here

https://github.com/jgbos/KFuddles

I go a little further with the module to allow the macro `@sequence` to automatically build the vector of weights and naming of parameters for saving.  It's a bit hacky for how I deal with indexing the weights though.  Feel free to use any of the code if it's useful.

CarloLucibello:
> When it comes to modules I don't know if you'd be interested in what I have been playing with here
> 
> https://github.com/jgbos/KFuddles

That is some really nice work! As much as I love the transparency of Knet, I would also like for it to offer that kind of higher level interface, has most deep learning frameworks do. Maybe @denizyuret has already some plans for that or would be interested in that sort of desing discussion?

jgbos:
yes, i think there is a case for building the weights and functions without high level interfaces.  But I also utilize the `Sequential` function in `PyTorch` all the time, so good to have both I think?

CarloLucibello:
also, working with  layer types, who can keep an internal state, as for instance is needed in batch normalization, makes for cleaner code where you do not have to pass parameters all around, as I do with `bmom` in this PR.

jgbos:
yeah, batch normalization was the main reason I started my code, the same reason you built `BatchMoments`.   I also wanted to way to automatically build the vector of weights and save parameters with unique names.

denizyuret:
I plan to work on a standard module interface ala #152 for 0.8.6.  I will look at this in the context of the whole modular style.

denizyuret:
I decided to work on better CUDNN integration for 0.8.6.  In particular the general CNN/RNN speed went up considerably. Please see #193 for @cgumeli's integration of the CUDNN batchnorm, we are still ironing out the interface. I have not benchmarked but suspect it will beat manual implementations. We will also look at dropout and softmax from CUDNN and replace existing implementations if they offer significant performance boost.

PR #146, opened 2017-07-12
[WIP] Making Knet Smaller

MikeInnes:
As @denizyuret and I discussed at length, it would be great for Knet to be a bit more modular. Having the KnetArray available separately would allow non-DL packages to take advantage of its functionality, which in turn will lead to more contributions and a lesser maintenance burden for Knet itself.

This patch strips out that array functionality and moves it to [KArrays.jl](https://github.com/MikeInnes/KArrays.jl), which Knet would depend on. Still early days, but the longer-term hope is that Knet shouldn't have to know about the GPU array type at all, and can get good performance with any AbstractArray.

Next steps for KArrays:

* 0.6 Support
* Factor out AutoGrad integration
* Factor out custom memory management, so any array type can use it.
* Rewrite CUDA kernels in pure Julia ([CUDAnative.jl](https://github.com/JuliaGPU/CUDAnative.jl)) for a simpler build step and maintenance.

Any thoughts on this setup? Obviously this isn't mergeable as is, but it'd be good to have some plan for how we can get it into Knet master; otherwise it will become too difficult to maintain the various branches.


jekbradbury:
Not sure where this should go (was probably just going to email Mike but I think this works too)
I spent a little time trying to see if I could wrap [DLPack](https://github.com/dmlc/dlpack)'s DLTensor type in order to automatically support Torch/MXNet/etc. operators in GPUArrays:
1. It looks like my idea of making that struct an isbits type won't work because the size and stride are represented by C arrays rather than inline (i.e. as Julia tuples).
2. Using an actually-isbits Julia struct for tensor metadata also won't work because of CUDA/CUDAnative argument size restrictions. 
3. Even in CUDA C, it still wouldn't be a good idea because nvcc compiles device kernels that take a large struct as an argument in such a way that the struct is copied into thread-local memory in the individual compute units, which slows things down a lot.
4. Even if we used generated functions and somehow considered the strides part of the tensor's type rather than passing them as an argument, it's not clear that making the tensors' indextype cartesian (in order to make 2/3/4-D tensors line up with the up-to-6D blockIdx and threadIdx system) is a good idea, because there are LOTS of gotchas with blockIdx and threadIdx.
So it's been a learning experience, and ultimately it looks like the approach that's been taken so far in all the Julia GPU projects (separate compact and strided tensor types; linear rather than cartesian indexing and device-inlined sub2ind etc. methods; specialized CUDA kernels wrapped in more generic dispatching host functions) is the best one. The three next steps Mike lists sound great, and I'm interested in helping with them.

What I definitely do want from GPUArrays in the long term is support for general `broadcast` and strided/subarrays (ideally views by default). It would be really cool to eventually support fusion of broadcast and reduction, and integrate with XDiff (e.g. https://github.com/dfdx/Espresso.jl/issues/16) for symbolic tensor expression differentiation and codegen...

MikeInnes:
Interesting research. Good in a way as I've been thinking along similar lines; with CUDAnative there's no real overhead to supporting strided arrays vs dense ones (just overload indexing). Where we want to reuse kernels (which may be relatively little if we can do it much better) it should be pretty easy to have a lightweight conversion between `StridedArray{<:CuArray}` and `DLTensor`.

In other words we can do the simplest, most natural thing and it should work out ok. We may well find that the design converges towards something like DLPack anyway, but we definitely shouldn't feel limited by it.

MikeInnes:
Updating Knet for 0.6 turned out to be a big rabbit hole, and I figured it was easier to start from scratch with a Julian approach. Hence [CuArrays](https://github.com/MikeInnes/CuArrays.jl), which supports the same functionality in a fully general way. I think the main missing pieces right now are more fleshed out cublas and cudnn integration.

Will close this for now, but would certainly appreciate feedback from the Knet side on what else is needed.

denizyuret:
I am deep inside the Julia 0.6 rabbit hole right now, hopefully will emerge soon. 
For CuArrays you mention cublas and cudnn.  Do we have full indexing, broadcasting, and reduction working?

MikeInnes:
Yes, those things should work in a fully general way, along with most other major array operations like cat and permute dims. Would be awesome if you could kick the tires on it, and I'll happily add whatever else is needed.

denizyuret:
@ilkerkesen was looking into GPUArrays, should we abandon that and switch
testing to CuArray?

ilkerkesen:
Last time I checked CuArray, it was a kind of KnetArray fork but now uses CUDAnative and it has a much more cleaner code base than KnetArray. GPUArrays was missing indexing and I didn't open an issue for that because it was already in TODO list in README.md. I guess indexing is still missing in GPUArrays. Before we move on, let me compile Julia v0.6 and test all of them again.

denizyuret:
@ilkerkesen @mikeinnes most of the code under Knet/test that runs with
Pkg.test("Knet") tests KnetArrays and can be used to test CuArrays.

Issue #144, opened 2017-07-11
Modular Interface


MikeInnes:
Knet's current API requires that all model parameters are collected together and passed into the model at once. This design has several issues: it makes code reuse more difficult, leading to hand-coded implementations of standard layers like `LSTM` in each model. It also makes it impossible to abstract over structure, so that (for example) replacing `LSTM` with `GRU` in a model requires significant refactoring rather than being a one-line change. Furthermore, it hampers performance: standard layers like `LSTM` cannot take advantage of the advanced static optimisations that are possible in other frameworks.

I want to discuss and get feedback on a new design that attempts to solve these problems without compromising Knet's simplicity and flexibility.

## AutoGrad Variables

Using a layer shouldn't require knowledge of its internals, so layers need to be able to create and manage their own parameters. In other words, tracked variables need to be part of the autograd API, rather than being hidden inside the `grad` function. Here's a sketch of what this could look like:

```julia
x = AutoGrad.Variable([1,2,3])
y = sum(2*x) # y::AutoGrad.Variable
back!(y, 1)
x.grad == [1, 1, 1]
```

A `Variable` is essentially the current `Rec`; it stores at least a value, a gradient, and the operation / inputs that created it. The `back!` function performs backpropagation and updates the gradients of input variables recursively, so `x` now stores a gradient that we can use for training.

## Layers

A layer is now just a function with some differentiable parameters.

```julia
struct Affine
  W
  b
end

Affine(in, out) = Affine(Variable(randn(in, out)),
                         Variable(randn( 1, out)))

(a::Affine)(x) = x*W + b
```

We can construct an `Affine` layer and then call it like a function on a normal array. Because the params are `Variable`s, the output will also be a `Variable` which we can use for backpropagation. This design makes Keras-like abstractions pretty trivial:

```julia
struct Chain
  layers::Vector{Any}
end

(c::Chain)(x) = foldl((c, x) -> c(x), x, c.layers)

Chain(Affine(10, 20), Affine(20, 5))
```

This method of defining layers is not at all incompatible with Knet's current approach (via `grad`); one can freely mix and match plain Julia functions with layers defined in this way.

## Training

Training does not need to change significantly; the only real difference is that in addition to the parameters and gradients supplied by `grad`, you'll have ones internal to the model. We can make it easy to pull these out by defining a common interface like `params` (e.g. `params(a::Affine) = [a.W, a.b]`); then you can train these variables in the obvious way.

[...]

PR #152, opened 2017-08-16
Modular Interface

ylxdzsw:
This is my implementation of the [modular interface](https://github.com/denizyuret/Knet.jl/issues/144) proposed by @MikeInnes.

Basically it exposes some AutoGrad logics by `track` and `back!`, `track(x)` creates a new tape and `track(x, y)` records `x` on the same tape as `y`. `back!` then run the backward pass and leave all gradients on the tape, which can then be retrieved by `getgrad`.

A model is a struct or closure which can be called. If it contains trainable parameters, it should track them in the forward pass so that them can be retrieved by `params`. If the input is not tracked, all parameters will also be "untracked" and the whole model is running in "prediction mode". In this case, `back!` is invalid.

To make code clearer for reviewing, it supports only Julia v0.6 now. We can add compat code later.

Any suggestions?

### DIFF
--- a/examples/charlm_using_modular_interface.jl
+++ b/examples/charlm_using_modular_interface.jl
@@ -1,0 +1,89 @@
+VERSION < v"0.6-" && error("currenctly modular interface only works for julia v0.6+")
+
+using Knet
+
+# load data
+
+const data = readstring(Knet.dir("data","10.txt"));
+const dict = unique(data);
+
+init(x...) = .2rand(x...) .- .1
+
+# define model
+
+const model = let
+    encoder = Embedding(length(dict), 64, init=init)
+    lstm1   = LSTM(64, 256, init=init)
+    lstm2   = LSTM(256, 256, init=init)
+    decoder = Affine(256, length(dict), init=init)
+
+    function CharLM(x, h)
+        h1, c1, h2, c2 = h
+        
+        input  = encoder(x)
+        h1, c1 = lstm1(input, h1, c1)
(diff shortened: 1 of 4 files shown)


denizyuret:
I will look at this for 0.8.6 release.

ngphuoc:
I've tried this pull request again and it works for me. However I notice the memory usage increases rapidly, roughly 4 times more than normal. I think the following code snippet allocates new memory every minibatch. Any idea we could optimize this a bit (reuse allocated memory for example)? 

```julia
"""
tracked_x = track(x)
track an input, `x` can be tuple, array or dict. If x is already tracked, it will be replaced with a new tape.
tracked_p = track(p, x)
track p on the same tape of x, return tracked p. If x is not tracked, it will also untrack p.
"""
track(x::AutoGrad.Rec) = track(x.value)
track(x, tape::AutoGrad.Tape=AutoGrad.Tape()) = AutoGrad.Rec(x, tape)
track(x, y) = x
track(x, y::AutoGrad.Rec) = AutoGrad.Rec(x, y.tapes[])
track(x::AutoGrad.Rec, y) = x.value
track(x::AutoGrad.Rec, y::AutoGrad.Rec) = x in y.tapes[] ? x : track(x.value, y)
```

denizyuret:
Please check out issue #347.

ylxdzsw:
The callable object approach in v1.1.0 is really nice. This one can be closed.

PR #153, opened 2017-08-22
Argument for saving generated texts to file is added to CharLM example

DoguD:
`—sresult generated.txt` argument  when used with `—load` and
`—generate` saves the generated text into a file.


denizyuret:
Thanks.

PR #154, opened 2017-08-24
compatibility with ubuntu 16.04

pvtmert:
quick and dirty fix for ubuntu 16.04 and cuda-toolkit (one from the repositories of ubuntu)


denizyuret:
Thank you.  Can you let me know what problem this fixes?  Is the change compatible with other platforms and compiler versions?

pvtmert:
hi there, i dont have a access to nvidia machine curently but
after installing `nvidia-{smi,modprobe,driver,cuda-toolkit,xconfig}` either install julia from default repositories or from tar archive, building fails because of some functions switched to `inline` which makes compiler to fail building

i will attach logs/screenshots when i get to ubuntu 16.04/nvidia machine

denizyuret:
This does not seem to be a problem in latest Ubuntu releases (I built successfully with Lubuntu 18). Can any other Ubuntu users confirm?

pvtmert:
i can confirm it has been fixed in upstream side (nvcc now implicitly assumes inline macro)

PR #160, opened 2017-09-19
Add fashion-mnist example

quaertym:
Fashion-MNIST is a new dataset that is designed to be a drop-in replacement for MNIST which I saw from LeCun’s post. I adapted MNIST example to work with this dataset. Here are [paper](https://arxiv.org/abs/1708.07747) and [repo](https://github.com/zalandoresearch/fashion-mnist). I thought classifying fashion products might be an interesting example. 

I replicated MLP score(0.877) reported on the [site](http://fashion-mnist.s3-website.eu-central-1.amazonaws.com/). 0.882 accuracy is achieved by runnning:
```
julia fashion-mnist.jl --hidden 100 --epochs 30
```

### DIFF
--- a/examples/fashion-mnist.jl
+++ b/examples/fashion-mnist.jl
@@ -1,0 +1,186 @@
+for p in ("Knet","ArgParse","Compat","GZip")
+    Pkg.installed(p) == nothing && Pkg.add(p)
+end
+
+"""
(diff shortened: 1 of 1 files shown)


denizyuret:
Thank you.

PR #161, opened 2017-09-19
Update vgg.jl for newer version of Images pkg

quaertym:


denizyuret:
Do you remember what this fixes?  I seem to get no error in Julia6 with the latest master.

quaertym:
Actually, this fixes the example for Julia5. It looks like `Pkg.add` does not install the latest version of `Images` when using Julia5. I think the reason for that is that more recent versions of `Images` depend on Julia6 features.

denizyuret:
OK, the problem was not julia6 or julia5 but julia4.  I fixed it with a conditional line.

quaertym:
OK, fix is 10953a5ff6cebabf11aa79a4f383a2d8cbefc779 for future reference. I am not a fan of version checks though.

PR #163, opened 2017-09-22
add back error functions

CarloLucibello:
This is the complementary part of denizyuret/AutoGrad.jl#31

Recent Knet/AutoGrad changes broke a lot of code involving error functions that I use in my research, so I would be very glad to have those back!


denizyuret:
I think the latest master fixes this, can you confirm?

CarloLucibello:
confirm

Issue #128, opened 2017-05-22
Package is not precompile-safe


tkelman:
As discussed here - https://github.com/denizyuret/Knet.jl/commit/ed076803c292511764765a6b057559fdc8d1a980

Due to the conditional dependency on JLD there, if JLD either becomes available after the Knet is precompiled or is removed after Knet has been precompiled, then you get either missing functionality or a failure to load the package. Most prominent on Julia 0.4, but the lack of cache tracking of conditional dependencies is an issue on all Julia versions. If you have to resort to manually managing the cache, then you aren't satisfying the preconditions to mark the package as `__precompile__(true)`.

```
~/Julia$ export JULIA_PKGDIR=$HOME/Julia/jlpkgtmp
~/Julia$ rm -rf $JULIA_PKGDIR
~/Julia$ julia-0.4/julia
               _
   _       _ _(_)_     |  A fresh approach to technical computing
  (_)     | (_) (_)    |  Documentation: http://docs.julialang.org
   _ _   _| |_  __ _   |  Type "?help" for help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 0.4.8-pre+3 (2016-11-04 23:38 UTC)
 _/ |\__'_|_|_|\__'_|  |  Commit 63b33a2 (198 days old release-0.4)
|__/                   |  x86_64-linux-gnu

julia> Pkg.init(); Pkg.add("JLD"); Pkg.add("Knet"); run(`$JULIA_HOME/julia -e 'using Knet'`); Pkg.rm("JLD"); using Knet
INFO: Initializing package repository /home/tkelman2/Julia/jlpkgtmp/v0.4
INFO: Cloning METADATA from git://github.com/JuliaLang/METADATA.jl
INFO: Cloning cache of BinDeps from git://github.com/JuliaLang/BinDeps.jl.git
INFO: Cloning cache of Blosc from git://github.com/stevengj/Blosc.jl.git
INFO: Cloning cache of Compat from git://github.com/JuliaLang/Compat.jl.git
INFO: Cloning cache of FileIO from git://github.com/JuliaIO/FileIO.jl.git
INFO: Cloning cache of HDF5 from git://github.com/JuliaIO/HDF5.jl.git
INFO: Cloning cache of JLD from git://github.com/JuliaIO/JLD.jl.git
INFO: Cloning cache of LegacyStrings from https://github.com/JuliaArchive/LegacyStrings.jl.git
[...]

PR #172, opened 2017-10-05
Use requires for conditional JLD dependency

MikeInnes:
This is much more robust to changes in package state, as well as helping to support the package in deployments (e.g. JuliaBox and JuliaPro).


CarloLucibello:
you may also want to take care of the SpecialFunctions.jl conditional dependence added in commit 557ffbd97378b5fb8192009bea4e4b57d584ccc8

MikeInnes:
That would be a good idea, although it's not that clear to me how that patch works; where are the functions being imported and overloaded?

CarloLucibello:
mhmh, I don't know. In julia 0.6 erf is still exported by Base, altough deprecated. 

In AutoGrad a similar conditional dependence was also made, but there you have
```julia
+if VERSION >= v"0.6.0" && Pkg.installed("SpecialFunctions") != nothing
-    Pkg.add("SpecialFunctions")
     eval(Expr(:using,:SpecialFunctions))
 end
```

denizyuret:
In Knet the conditional stuff is in:
https://github.com/denizyuret/Knet.jl/blob/master/src/unary.jl#L65

Just grep for SpecialFunctions in AutoGrad and/or Knet...

MikeInnes:
It's slightly tricky to combine the requires block with the version check in this case. I may do that as a separate PR.

denizyuret:
I get the following error in Julia4, is this easy to fix?
```
ERROR: LoadError: UndefVarError: @init not defined
```

denizyuret:
Just realized @MikeInnes's package only supports Julia 6+.  Seems to not complain in Julia 5 but not sure if it is functional.  I could do a conditional thing where I have the old code for Julia < 6 and Mike's code for >= 6.  Will that fix https://github.com/denizyuret/Knet.jl/issues/128 ?

MikeInnes:
Requires master only supports 0.6, but there are versions for 0.4 and 0.5 as well. 0.4 is probably stale at this point but may be easy to fix.

> I could do a conditional thing where I have the old code for Julia < 6 and Mike's code for >= 6. Will that fix #128 ?

For version 0.6, yes. That would be fine by me.

denizyuret:
Hm, at some point you decided not to export @init in julia4, any reason?

MikeInnes:
Honestly, no idea. If you add it back and it doesn't cause issues I can re-tag for 0.4.

denizyuret:
@MikeInnes do conditional dependencies work in Julia 1.0?  It seems like if a dependency is not in Project.toml we get a big warning even if it is installed asking why we are using a package we did not declare as a requirement...

PR #174, opened 2017-10-12
add Nesterov optimizer

### LINKED ISSUE #169: Missing Nesterov momentum
It should be fairly easy to implemet as a slight variation of the standard momentum, using eq. 7   in https://arxiv.org/pdf/1212.0901.pdf (simplified Nesterov momentum)

CarloLucibello:
fixes #169


CarloLucibello:
this failure on the gpu builbots has been going on for some time, what is causing it?

ilkerkesen:
Both Julia v0.5 and v0.6 builds had the same problem,

```
Cannot find libknet8, please rerun Pkg.build("Knet").
```

denizyuret:
Thank you.

PR #178, opened 2017-10-20
Use CUDAapi and CUDAdrv to figure out build flags

maleadt:
This requires https://github.com/JuliaGPU/CUDAapi.jl/pull/6 and a tag.

cc @denizyuret


denizyuret:
Does CUDAapi and CUDAdrv work ok on non-gpu machines with no libraries?

MikeInnes:
CUDAapi is fine, but CUDAdrv expects CUDA to be available and will give build errors if it's not installed. I think that behaviour is correct, but it also means you shouldn't even try to install it on non-GPU machines.

My solution in Flux has been to split the CUDA array type from the ML parts (which just work with AbstractArray). That would be more work, but it's also a nice refactoring in that it makes it trivial to test across array types (new/old cuda-backed arrays, clarrays etc).

maleadt:
In addition, CUDAdrv failing to build is nasty but non fatal. The package will also load properly, but using its functionality will fail and should be guarded by a check against `CUDAdrv.configured`.

maleadt:
Any update here, or should I disable CI in the meantime?
FYI, I can configure my CI to have certain packages pre-installed before installing Knet, so you could also check `Pkg.installed` in `build.jl` (as an alternative to having CUDAapi/CUDAdrv in REQUIRE and checking the package's `configured` flag, which integrates better with eg. `Pkg.resolve` but leads to nasty but harmless build errors when installing CUDA*.jl on systems without CUDA).

denizyuret:
I think checking Pkg.installed is a good idea -- getting errors on cpu
systems is confusing to users.  In that case I would only need CUDAapi and
CUDAdrv preinstalled before Knet I guess? Will you configure this or is
there something I can do?  Right now I am trying to figure out a way to
make things work on a windows machine -- it may work out of the box, or I
may have to send some PRs to CUDAapi.  As soon as I figure that out, I will
start using the CI -- so no need to disable unless I am taking up some
resources right now.

best,
deniz

maleadt:
Added CUDAdrv (and as a result CUDAapi) as prerequisite: https://github.com/maleadt/juliagpu-buildbot/commit/92d26f314f6adea86bb3eaba8e4d5482ca821fc4

denizyuret:
I think toolchain.flags in Knet/deps/build.jl have been deprecated. What do I replace it with?

I also see CUDAdrv is only used to discover --gpu-architecture. What is the downside of not specifying architecture? (possibly to eliminate dependency on CUDAdrv).

Finally, do you have any advice on Windows?  Will CUDAapi discover the right compiler and options? Will Libdl.find_library work cross platform?

maleadt:
> I think toolchain.flags in Knet/deps/build.jl have been deprecated. What do I replace it with?

Those only contained `--compiler-bindir`, just specify that manually. eg: https://github.com/JuliaGPU/CUDArt.jl/blob/4134b94d184df06bfb77f546e532a99b6c9f0493/deps/build.jl#L59-L60

> I also see CUDAdrv is only used to discover --gpu-architecture. What is the downside of not specifying architecture? (possibly to eliminate dependency on CUDAdrv).

Possible slower code. You can drop it if you like, you weren't using it before.

> Finally, do you have any advice on Windows? Will CUDAapi discover the right compiler and options? Will Libdl.find_library work cross platform?

Probably not, I don't have a working Windows+CUDA set-up, but I'd welcome patches or issue reports to improve that situation.

denizyuret:
We can remove the Julia 0.5 CI for Knet.  I am dropping support for pre-0.6
versions.  Julia-dev is currently failing but I will fix that at some point.

maleadt:
OK, done.

PR #188, opened 2017-10-30
Using _sync intrinsic for newer CUDA versions (9.0+)

ddikbayir:
Added conditional pre-processor macro to multidimensional broadcast kernel (cuda13) to support _sync intrinsic introduced with CUDA version 9.0.


denizyuret:
Integrated in master, thanks.

PR #189, opened 2017-11-07
Cudnn

cangumeli:


denizyuret:
TD is broken, code cannot be run in general.  Too many files are changed, I cannot test.  I recommend the following:
* Let us leave refactoring for later.
* Start with current Knet/master and just add rnn.jl.
* Don't change any existing code or type, just add new ones as needed.
* That way we can be sure nothing else breaks and things are easy to merge.
* After we can get RNN to work we write unit tests and documentation.
* After all is complete we can worry about refactoring.

PR #193, opened 2017-11-14
Batchnorm(CUDNN + CPU) Work in Progress

cangumeli:
I created GPU and CPU primitives for 4d spatial batch normalization, which is the one used in deep CNNs. 2d batch norm is implemented using 4d version by padding two dimensions from the start and unpacking them. Both versions seem to work correctly and pass grad checks; but I still need to check some corner cases. I will document and discuss the API as soon as possible. 

After the relevant work is done, we should test the functions further, especially by training real deep MLPs and CNNs.


cangumeli:
I finalized the batchnorm API and added a working CIFAR 10 example. Documentation for the public API can be found in batchnorm.jl file.  I also implemented several unit tests that can be extended.

The batchnorm function has four methods:

`batchnorm(g, b, x; kwargs...)`
`batchnorm(x; kwargs...)`
`batchnorm(m::BNMoments, g, b, x; kwargs...)`
`batchnorm(m::BNMoments, x; kwargs...)`

`g` and `b` are affine parameters and `x` is the input. Without `g` and `b`,`batchnorm` just normalizes `x` to a unit gaussian. Training/testing separation is accomplished by a boolean keyword argument named `training`, whose default value is `true` and controlled by the user. 

BNMoments is a data structure for storing running mean and running variance in fields `mean` and `var`. You can simply initialize it using `BNMoments()`  without any argument in most use cases. In the methods of `batchnorm` moments is not the first argument, there is a keyword argument called `moments` which is `nothing` by default and can be `nothing` when `training` is true. This flexibility is given to the users to support different approximation methods. For example, one may prefer sampling a large mini-batch of data and running `batchnorm` once such that:
```
m = BNMoments(;momentum=1)
batchnorm(g, b, large_batch_of_x; training=true, moments=m)
```

Further details about `batchnorm` are in the documentation and the CNN example.

denizyuret:
Some comments, suggestions...

On Fri, Nov 24, 2017 at 12:24 AM cangumeli <notifications@github.com> wrote:

> I finalized the batchnorm API and added a working CIFAR 10 example.
> Documentation for the public API can be found in batchnorm.jl file. I also
> implemented several unit tests that can be extended.
>
> The batchnorm function has four methods:
>
Does the user need all four methods, suggestions below...

> batchnorm(g, b, x; kwargs...)
> batchnorm(x; kwargs...)
>

Could we combine these two methods by making x the first argument and g,b
the optional 2nd and 3rd arguments?

>
> batchnorm(m::BNMoments, g, b, x; kwargs...)
> batchnorm(m::BNMoments, x; kwargs...)
>
Could we combine these with the previous two by making m a keyword argument
with a default value?

> g and b are affine parameters and x is the input. Without g and b,
> batchnorm just normalizes x to a unit gaussian. Training/testing
> separation is accomplished by a boolean keyword argument named training,
> whose default value is true and controlled by the user.
>
Why not handle `training` automatically like we now do in rnnforw and
dropout?

> BNMoments is a data structure for storing running mean and running
> variance in fields mean and var. You can simply initialize it using
> BNMoments() without any argument in most use cases. In the methods of
> batchnorm moments is not the first argument, there is a keyword argument
> called moments which is nothing by default and can be nothing when
> training is true. This flexibility is given to the users to support
> different approximation methods. For example, one may prefer sampling a
> large mini-batch of data and running batchnorm once such that:
>
> m = BNMoments(;momentum=1)
> batchnorm(g, b, large_batch_of_x; training=true, moments=m)
>
>
Could we give the same flexibility without 4 separate methods?

cangumeli:
> Could we combine these two methods by making x the first argument and g,b
the optional 2nd and 3rd arguments?

Yes we can, it can be easily done by just changing the argument arrangement of the `batchnorm` function, keeping the internal logic. I did it that way to follow the Knet convention where learnable weights are first arguments, but I can change it by editing just two lines of code.

> Could we combine these with the previous two by making m a keyword argument
with a default value?

I don't think it is convenient, because the same `BNMoments` object is needed for each call of the same `batchnorm`. Furthermore, user will need to serialize the running mean and running var together with the model weights when backing up a model, so user needs full access to the moments object. 

> Why not handle `training` automatically like we now do in rnnforw and
dropout?

In `dropout` and `rnnforw`, we assume the operation is in the training mode when arguments are recorded. Training mode of `batchnorm` means we are using mean and variance of the `x` and updating the `moments`, rather than using the mean and variance stored in the `moments`. User may want to run `batchnorm` in testing mode while training a batch-normalized model,  or call `batchnorm` in training mode just to estimate `moments`. An example of first use case is the [Resnet Faster-RCNN model](https://arxiv.org/abs/1512.03385), while an example of second is the large minibatch example I gave. There is even a test mode backward I implemented for the first use case in `batchnorm_back` function.

However, we can change the name of the keyword argument to something like `train`, if you think it will cause any confusion.


In conclusion, I can keep just two methods:
`batchnorm(x, g=nothing, b=nothing; o...)`
`batchnorm(m::BNMoments, x, g=nothing, b=nothing; o...)`

An alternative to this is keeping only the method taking moments as the first argument, which will be the one used most of the time. However, this would introduce dummy computation when user wants to use other estimation methods for mean and var.

examples/cifar10-cnn/cnn_batchnorm.jl
```diff
@@ -0,0 +1,123 @@
+using Knet
+include(Pkg.dir("Knet", "data", "cifar.jl"))
+
+function loaddata()
+    info("Loading CIFAR 10...")
+    xtrn, ytrn, xtst, ytst, = cifar10()
+    mn = mean(xtrn, (1,2,4))
+    xtrn = xtrn .- mn
+    xtst = xtst .- mn
+    ytrn = onehot(Int.(ytrn))
```
denizyuret: Do you need to use onehot?  We are using nll now, which takes integer answers.
Also if this mean subtraction thing is standard I can make it part of the cifar10() function.

examples/cifar10-cnn/cnn_batchnorm.jl
```diff
@@ -0,0 +1,123 @@
...
+    w = Any[
+        kaiming(et, 3, 3, 3, 16), bnparam(et, 16),
+        kaiming(et, 3, 3, 16, 32), bnparam(et, 32),
+        kaiming(et, 3, 3, 32, 64), bnparam(et, 64),
+        xavier(et, 100, 8 * 8 * 64), bnparam(et, 100),
+        xavier(et, 10, 100), zeros(et, 10, 1)
+    ]
+    # Initialize a BNMoments object for each batchnorm
+    m = Any[BNMoments() for i = 1:4]
+    if gpu() >= 0; w = map(KnetArray{et}, w); end
```
denizyuret: How do you keep track of the alignment between w and m?

examples/cifar10-cnn/cnn_batchnorm.jl
```diff
@@ -0,0 +1,123 @@
...
+    if maxpool; o=pool(o); end
+    return o
+end
+
+function lin_layer(w, m, x)
+    o = w[1] * x
+    o = batchnorm(o, m, w[2])
+    return relu.(o)
+end
+
```
denizyuret: These layers are really nice and easy to understand.

examples/cifar10-cnn/cnn_batchnorm.jl
```diff
@@ -0,0 +1,123 @@
...
+lossgrad = grad(loss)
+
+# Training
+function epoch!(w, m, p, xtrn, ytrn;  mbatch=64, wdecay=1e-5)
+    N = size(ytrn)[end]
+    # Shuffling data is recommended when using batch norm
+    indices = shuffle(1:N)
+    DT = gpu() >= 0 ? KnetArray : Array
```
denizyuret: You check for gpu() in multiple places, maybe have a single atype constant at the beginning that can be KnetArray{Float32} or Array{Float32} based on gpu()?

examples/cifar10-cnn/cnn_batchnorm.jl
```diff
@@ -0,0 +1,123 @@
...
+    for i = 1:mbatch:(N - N % mbatch) #ignore remainder for consistent batch sizes
+        ind = indices[i:i+mbatch-1]
+        x = DT(xtrn[:, :, :, ind])
+        y = DT(ytrn[:, ind])
+        g = lossgrad(w, m, x, y)
+        if wdecay>0; g = g .+ wdecay .* w; end
```
denizyuret: What is wdecay? It looks a bit like momentum?

examples/cifar10-cnn/cnn_batchnorm.jl
```diff
@@ -0,0 +1,123 @@
...
+        for (wi, gi, pi) in zip(w, g, p)
+            update!(wi, gi, pi) 
+        end
+    end
+end
+
+function acc(w, m, xtst, ytst; mbatch=64)
```
denizyuret: I think this can be handled by the new accuracy() function if you use integer answers and the built-in minibatch().

examples/cifar10-cnn/cnn_batchnorm.jl
```diff
@@ -0,0 +1,123 @@
...
+end
+
+function train(;epochs=10)
+    w, m = init_model()
+    o = map(_->Adam(), w)
+    (xtrn, ytrn), (xtst, ytst) = loaddata()   
+    for epoch = 1:epochs
+        println("epoch: ", epoch)
+        epoch!(w, m, o, xtrn, ytrn)
+        println("train accuracy: ", acc(w, m, xtrn, ytrn))
```
denizyuret: Train accuracy here may be expensive, you could use gradloss() instead of grad() and report loss for free.

examples/cifar10-cnn/cnn_batchnorm.jl
```diff
@@ -0,0 +1,123 @@
...
+        println("test accuracy: ", acc(w, m, xtst, ytst))
+    end
+end
+
+# TODO: add command line options
+function main(ARGS=nothing)
+    train()
+end
+
+main()
```
denizyuret: We need documentation or better yet convert this into a notebook...

src/Knet.jl
```diff
@@ -26,6 +26,7 @@ include("broadcast.jl");        # elementwise broadcasting operations
 include("reduction.jl");        # sum, max, mean, etc.
 include("linalg.jl");           export mat # matmul, axpy!, transpose, (i)permutedims
 include("conv.jl");             export conv4, pool, deconv4, unpool
+include("batchnorm.jl");        export batchnorm, BNMoments, bnparam, bnscale, bnbias
```
denizyuret: Use a regular lowercase bnmoments() function instead, in case we change our minds about using a struct here in the future?

src/batchnorm.jl
```diff
@@ -0,0 +1,485 @@
+"""
+
+`BNMoments` is a high-level data structure used to store running mean and running variance
+of batch normalization.
+
+# Fields
+ `momentum::AbstractFloat`: A real number between 0 and 1 to be used as the scale of
```
denizyuret: This is counterintuitive: normally high momentum means you preserve your old values more, no?

src/batchnorm.jl
```diff
@@ -0,0 +1,485 @@
...
+ 
+ `mean`: The running mean.
+
+ `var`: The running variance.
+ 
+ `meaninit`: The function used for initialize the running mean. Should either be `nothing` or
+of the form `(eltype, dims...) -> data`. `zeros` is a good option.
+
+ `varinit`: The function used for initialize the running variance. Should either be `nothing` or
+`(eltype, dims...) -> data`. `ones` is a good option.
```
denizyuret: Doesn't it take a long time to converge to actual values if we use ones and zeros?  I don't know if this is a problem in practice, but if it is we could try to set the mean and variance from the first few actual examples.

src/batchnorm.jl
```diff
@@ -0,0 +1,485 @@
...
+
+# Constructors
+ `BNMoments(;momentum=0.1, meaninit=zeros, varinit=ones)` sets `mean` and `var` to nothing.
+When this constructor is used, `batchnorm` functions will initialize `mean` and `var` during
+the first training iteration to have relevant dimensions and match the data type of the input.
+
+ `BNMoments(momentum, mean, var)` can be used directly load moments from data. `meaninit` and
```
denizyuret: OK, so if I don't know anything about mean/var I just use bnmoments() before training to initialize. It will learn the real mean/var.  During testing it will stop updating itself and use the final frozen values.  I only need bnmoments(momentum, mean, var) if I am doing something special and loading mean/var from external source.  Is this during training or testing?  If testing, why do I need momentum?  If training what is the difference between this and starting bnmoments with a different init?

src/batchnorm.jl
```diff
@@ -0,0 +1,485 @@
...
+BNMoments(;momentum=0.1, meaninit=zeros, varinit=ones, o...) =
+    BNMoments(momentum, nothing, nothing, meaninit, varinit)
+
+BNMoments(momentum, mean, var) = BNMoments(momentum, mean, var, nothing, nothing)
+
+
+#TODO: improve the documentation. Explain what batchnorm is.
+"""
+
+`batchnorm(x[, moments, param]; kwargs...)` performs batch normalization to `x`
```
denizyuret: call third arg params as it represents gamma and beta.

src/batchnorm.jl
```diff
@@ -0,0 +1,485 @@
...
+with optional scaling factor and bias stored in `w`. Operation performed is spatial
```
denizyuret: what is w?

src/batchnorm.jl
```diff
@@ -0,0 +1,485 @@
...
+
+BNMoments(momentum, mean, var) = BNMoments(momentum, mean, var, nothing, nothing)
+
+
+#TODO: improve the documentation. Explain what batchnorm is.
+"""
+
+`batchnorm(x[, moments, param]; kwargs...)` performs batch normalization to `x`
+with optional scaling factor and bias stored in `w`. Operation performed is spatial 
+batch normalization if x is 4d or 5d. `moments` is a `BNMoments` object that stores
```
denizyuret: Are there other options than 4d and 5d?  What is spatial batch normalization, are there other alternatives?

src/batchnorm.jl
```diff
@@ -0,0 +1,485 @@
...
+#TODO: improve the documentation. Explain what batchnorm is.
+"""
+
+`batchnorm(x[, moments, param]; kwargs...)` performs batch normalization to `x`
+with optional scaling factor and bias stored in `w`. Operation performed is spatial 
+batch normalization if x is 4d or 5d. `moments` is a `BNMoments` object that stores 
+running mean and variance to be used in testing. It is optional in training mode, 
+but mendatory in test mode. `param` stores the optional affine parameters gamma and beta.
+`bnparam` can be used to initialize `param`.
+
```
denizyuret: It would be nice to have usage examples in the doc.  In particular the most standard usage (where both train and test use a shared bnmoments object and learn bnparams) and less standard uses of providing bnmoments or bnparams externally. This should motivate why optional things are optional.

src/batchnorm.jl
```diff
@@ -0,0 +1,485 @@
...
+# Keywords
+
+ `eps=1e-5`: The epsilon parameter added to the variance to avoid division by 0
+ 
+ `training=nothing`: When `training` is true, the mean and variance of `x` are used and `moments`
```
denizyuret: Take a look at how this is implemented in dropout.  training is a boolean, its default is true for the x::Rec method and false for the generic x method.

src/batchnorm.jl
```diff
@@ -0,0 +1,485 @@
...
+ argument is modified if it is provided. When `training` is false, mean and variance stored in 
+ the `moments` argument are used. When training is kept `nothing`, the mode is determined
+ based on inputs' being recorded.
+
+"""
+function batchnorm(x, moments::Union{BNMoments, Void}=nothing, param=nothing;
+                   training=nothing, o...)
+    xnd = ndims(x)
```
denizyuret: Do we need o... for anything in keyword args?

src/Knet.jl
denizyuret: Also, why are we exporting bnscale and bnbias?

test/batchnorm.jl
```diff
@@ -0,0 +1,97 @@
...
+srand(42)
+TOL=0.1
+
+# CPU 4d buffers
+ax4 = Array(5rand(4,5,5,3));
+aw4 = bnparam(5)
+# GPU 4d buffers
+kax4 = KnetArray(5rand(4,5,5,3));
+kaw4 = KnetArray(bnparam(5));
+
```
denizyuret: Let's test all element types (Float32,Float64), array types (KnetArray,Array), and dimensions (2,4,5)?

test/batchnorm.jl
```diff
@@ -0,0 +1,97 @@
...
+# Array types
+at = typeof(ax4)
+kat = typeof(kax4)
+
+# utils
+stdev(x) = mean((x .- mean(x)) .* (x .- mean(x)))
```
denizyuret: 1. This seems like variance.
2. There should be a standard Julia function for this, var()?

examples/cifar10-cnn/cnn_batchnorm.jl
cangumeli: I used the old, familiar method to make things work quicker. `nll` will be more natural, so I'm planning to remove one-hot encoding.

examples/cifar10-cnn/cnn_batchnorm.jl
cangumeli: There are 4 batchnorm layers, so there should be a `BNMoments` object for each of them. In general, it is equal to `bnparam` calls. In more complex examples, user should initialize w and m arrays. This may be inconvenient for very deep models like resnet, but I can't think any solution to this problem until we switch to the modular API.

examples/cifar10-cnn/cnn_batchnorm.jl
cangumeli: It is supposed to be l2 decay if I was not mis-implemented it (each weight contributes to the error gradient of itself `decay .* w`). It doesn't help a a lot and seems confusing, so I think I should remove it.

src/Knet.jl
cangumeli: `bnmoments` seems a better name, since it matches with the `bnparams` conventionally. So  it is reasonable to make this change. Should user still know moments is a struct, or should we add accessor methods like `mean(moments)`, `var(moments)` etc.? 

I exported bnscale and bnbias for easy access to g and b, but I think I'll remove these exports, they just return first and second halves of the same array, without even reshaping.

src/batchnorm.jl
cangumeli: I just checked, the original name of this parameter in cudnn is `exponentialAverageFactor`.  Is it better if we call it something like `factor`?

src/batchnorm.jl
cangumeli: In batchnorm, people often assume everything is a unit gaussian initially. I saw `ones` is the default initialization in other frameworks like [TF](https://www.tensorflow.org/api_docs/python/tf/layers/batch_normalization) and [PyTorch](http://pytorch.org/docs/master/_modules/torch/nn/modules/batchnorm.html), so I followed this convention. I'm not sure making it `zeros` would change anything.

src/batchnorm.jl
cangumeli: I realized forcing to provide `momentum` when giving `mean` and `var` is problematic when loading pre-trained models just for inference. `meaninit` and `varinit` are only called when `mean` and `var` are nothing, so we may reduce `bnmoments` to a single function with default arguments.  

Calling `bnmoments(mean, var)` and `bnmoments(meaninit=(_)->mean, varinit=(_)->var)` are similar; but calling the second will provide automatic type conversion (for instance, if `mean` is an `Array` and `x` is  a`KnetArray`, mean will be made `KnetArray` and stored as a `KnetArray` in the `moments`). This behaviour is provided by the `_lazy_init!` helper function and works when `x` is a `Rec` or first call made when `training=false`. When `mean` and `var` are provided by the user, they are not further processed and used directly.

src/batchnorm.jl
cangumeli: In CNNs, it is assumed each channel is a single feature, so batch norm is performed by averaging the input in all dimensions other than the channel dimesion (dimension `ndims(x)-1`).  There is also a per-activation batch normalization, which averages all dimensions other than the batch dimension. The two modes are identical when the input is 2d. I didn't support per activation mode for 4d and 5d since it is not common and user can perform this without any overhead by calling `reshape(batchnorm(mat(x)), size(x))`.

The normalization dimension will be added to the doc. We also support 2d inputs, which I'll  ad to the documentation. I'm not sure what to do for 3d inputs, so I didn't support them for now.

src/batchnorm.jl
cangumeli: There were some options like `cache_verbose` that I used for debugging. There are also `cudnnhandle`, `alpha`, 'beta` etc. for the GPU version, which I did not documented. I added o... since `batchnorm` is a wrapper over primitives, so o... may make it easier to integrate more primitive keyword arguments to the public API. 

I don't think removing o... would break anything.

test/batchnorm.jl
cangumeli: 1.  Yes, stdev is not the correct name, so I'll change it. The stdev is only used for a sanity check  and the result should be 1, so test is still valid.
2.  There was no `var` function that takes `KnetArray` when the last time I merged with the upstream, so I performed this operation. I can make it more efficient when I increase the number of tests.

examples/cifar10-cnn/cnn_batchnorm.jl
cangumeli: Is it the same thing? The model is updated after each lossgrad call, so I thought we should freeze the model when computing accuracy.

examples/cifar10-cnn/cnn_batchnorm.jl
```diff
@@ -0,0 +1,139 @@
...
+# Training
+function epoch!(w, m, o, xtrn, ytrn;  mbatch=64)
+    data = minibatch(xtrn, ytrn, mbatch;
+                   shuffle=true,
+                   xtype=atype())
+    for (x, y) in data
+        g = lossgrad(w, m, x, y)
+        for (wi, gi, oi) in zip(w, g, o)
+            update!(wi, gi, oi) 
+        end
```
CarloLucibello: cannot do a single `update!(g, w, o)` here?

cangumeli: Yes, it can be done. I changed it to a single line update!.

examples/cifar10-cnn/cnn_batchnorm.jl
denizyuret: It is not the same thing, because as you said the train loss for the model at the beginning of the epoch will not be the same as at the end of the epoch.  However if the purpose is to get some progress indication it is good enough.  I don't know if these acc calls cost too much time so it may not be worth a change.

PR #208, opened 2017-12-01
reshape with colon for KnetArray

CarloLucibello:


denizyuret:
Thanks.

denizyuret:
There was no 'function' keyword in front of the new reshape. It broke my compile. How did it work in your test?

CarloLucibello:
Oh, that was a last second (supposedly) aestethic change of an already present method that I did for readability. Sorry for that, the fact that it was a trivial change made me careless

PR #210, opened 2017-12-02
Generic ResNet Implementation

cangumeli:
I re-implemented a more general version of resnet to be used in benchmark and research. This version use newly added cudnn-based `batchnorm` implementation, include cifar models together with imagenet models and can be initialized randomly without loading pre-trained weights.  I also support a keyword argument `stage` in both initialization and forward pass functions. For example, `stage=3` means you'll get 14x14 feature maps from the output of the third group of residual blocks.
I think we can use this version in benchmarks and then add a CIFAR training example to the folder.


cangumeli:
For now, we may keep both versions and the notebook. I didn't update the demo in the notebook yet.

I added a simple benchmark file adapted from 
https://github.com/ilkarman/DeepLearningFrameworks/tree/master/inference, which I used in CN2 (K80). Our ResNet-50 inference time seems to be about 8.2 seconds.

I need to add that there is some error in the benchmark file causing `MethodError(Core.Inference.convert, (AssertionError, "invalid age range update"), 0x0000000000000ac6)` to be thrown when I quit Julia REPL. This is not related to benchmark functions - it also happens when I just include the file and quit without running anything. I couldn't figure out what is the source of this error. This problem may solve itself when we make the whole thing  a notebook.

denizyuret:
Any idea why the CPU part is so slow?  Is there an operation that particularly slows things down that we can improve looking at pytorch, tf etc?

denizyuret:
I cleaned up a few things, implemented a single array version of minibatch and I tested on AWS and I got 8.05 for GPU which puts us at the top. Since this is a benchmark we should be careful we are running the exact same operations. How sure are you that small model differences will make no difference in timing?  How difficult is it to use the resnet everybody else is using in ilkarman's repo?

cangumeli:
> How sure are you that small model differences will make no difference in timing? How difficult is it to use the resnet everybody else is using in ilkarman's repo?

The benchmark seems to be about the standard `resnet` implementations provided by the frameworks. For instance, [Keras  placed strides in 1x1 convs](https://github.com/fchollet/keras/blob/master/keras/applications/resnet50.py#L103), while [pytorch  placed them in 3x3 convs](https://github.com/pytorch/vision/blob/master/torchvision/models/resnet.py#L64). Therefore, since using matconvnet models is more convenient than parsing torch weights etc., I think we can keep the current version. 

> Any idea why the CPU part is so slow? Is there an operation that particularly slows things down that we can improve looking at pytorch, tf etc?

The network mostly consists of `conv4`, `batchnorm` and `relu.`.  I'll be checking times of them in CPU and discuss the results.

cangumeli:
By the way, there are two possible issues about the [last version of our benchmark notebook](https://github.com/denizyuret/Knet.jl/blob/master/examples/resnet/resnet_benchmark.ipynb). 
1. Other benchmarks (tf and pytorch) copy the results into a CPU numpy buffer. I don't see any code copying the results to a CPU buffer in our notebook.
2. We automatically adjust operation modes (training, testing) in `batchnorm`. Other frameworks might not be adjusting the mode and this may slow them down. However, I guess this is an issue of benchmark.

denizyuret:
OK, I forgot about the cpu transfer.  I'll see if doing it before or after
the hcat is better.

cangumeli:
Current version of `batchnorm` is using standard vectorized Julia operations in CPU, which is slower than 1x1 `conv4s` with large channel sizes (convs in 1024-channel group of blocks in ResNet, which is often the deepest).

A re-implemented CPU `batchnorm` using Julia 6 fusions and opened a [PR.](https://github.com/denizyuret/Knet.jl/pull/216) In my laptop, fusions are several times faster for large channel sizes and perform less memory allocations. For further optimizations, we may rewrite the operation in C++ and OpenMP .

Also `conv4` uses im2col, so trying MKL might help large matrix multiplications, if we haven't used it in benchmarks already.

denizyuret:
Thanks!  I will try it out.  I haven't tried MKL. We should try it as thats
what the top cpu inference system uses.

denizyuret:
However MKL is not high priority right now, as nobody is going to run
resnet on cpu.

denizyuret:
Can you write a readme under resnet to tell users which file does what?

PR #216, opened 2017-12-06
Fused Batchnorm for CPU

cangumeli:
I re-implemented the batchnorm CPU using [Julia 6 fusion operator](https://julialang.org/blog/2017/01/moredots). CPU tests are passing. In my laptop, I obtained 2.5-5 x speedups in forward pass after a few cold-start executions. This might improve our CPU benchmarks in ResNet and other CNN models.


denizyuret:
I checked in the final implementation of the resnet_benchmark notebook in Knet and submitted a pull request to @ilkarman.

PR #223, opened 2017-12-10
CPU bidirectional RNN Implementation and Fixes

cangumeli:
I implemented the bidirectional RNN for CPU. Code is very creepy, but works consistently with the GPU version without breaking any test case.
 
Other important changes I've made:
1. Proper initalization in CPU using views .
2. No dependency to cuDNN when using CPU. I edited `handle` arguments and removed dependency to cuDNN functions (param size etc.) in CPU initialization.

I stored my terminal logs to show the consistent behaviour, but we should add automated tests, make the code more readable and consider what will happen when we try supporting `batchSizes`.


src/rnn.jl
```diff
@@ -159,32 +159,37 @@ function cudnnGetFilterNdDescriptor(wDesc::FD; nbDimsRequested = 8)
     end
 end
 
+
+# call gpu everytime to reflect device changes
+gethandle() = gpu() >= 0 ? cudnnhandle() : nothing
```
denizyuret: If this is how we want to handle cpu, should I make this the default behavior of cudnnhandle?

src/rnn.jl
```diff
@@ -235,6 +240,9 @@ function rnnparam(r::RNN, w, layer::Integer, id::Integer, par::Integer; handle=c
             for l = 1:layer, i = 1:ids
                 if inputLayer(r,l) && i <= inputIds
                     len = (r.inputMode==0 ? XH : 0)
+                elseif l>2 && r.direction==1 && i <= div(ids, 2)
+                    # bidirectional weight
+                    len = 2HH
```
denizyuret: Isn't this two separate weights, each HH in size, according to cudnn interface?

src/rnn.jl
```diff
@@ -159,32 +159,37 @@ function cudnnGetFilterNdDescriptor(wDesc::FD; nbDimsRequested = 8)
...
-* For bidirectional, the same applies to rnnparam(r,w,2,I,1): the first back layer.
+    The effect of skipInput: Let I=1 for RELU/TANH, 1:3 for GRU, 1:4 for LSTM
+    * For skipInput=false (default), rnnparam(r,w,1,I,1) is a (inputSize,hiddenSize) matrix.
+    * For skipInput=true, rnnparam(r,w,1,I,1) is `nothing`.
+    * For bidirectional, the same applies to rnnparam(r,w,2,I,1): the first back layer.
 
-"""
-function rnnparam(r::RNN, w, layer::Integer, id::Integer, par::Integer; handle=cudnnhandle())
+    """
+function rnnparam(r::RNN, w, layer::Integer, id::Integer, par::Integer; handle=gethandle(), useview=false)
```
denizyuret: Do we need useview to be an option?  We can just pick the more efficient implementation.

src/rnn.jl
```diff
@@ -251,38 +259,40 @@ function rnnparam(r::RNN, w, layer::Integer, id::Integer, par::Integer; handle=c
             i2 = i1 + len - 1
         end
     end
+    @inline access(a, rng) = (isa(a, KnetArray) || ~useview) ? a[rng] : view(a, rng)
```
denizyuret: I should implement view for KnetArray instead of forcing you to define a function here.  Of course only contiguous views will be supported and they will just call getindex, which is already implemented as view in KnetArray. I'll do this if we confirm that using views with Arrays really helps the speed (which may not, since the backward pass still needs to use separate arrays).

src/rnn.jl
```diff
@@ -371,9 +381,23 @@ function rnninit(inputSize, hiddenSize;
         w = KnetArray{dataType}(1,1,cudnnGetRNNParamsSize(r))
     else
         r = RNN(inputSize,hiddenSize,numLayers,dropout,inputMode,direction,mode,algo,dataType,nothing,nothing,nothing,nothing,nothing)
-        w = Array{dataType}(1,1,cudnnGetRNNParamsSize(r))
+        # TODO: make this a separate function?
```
denizyuret: But I already implemented this in cudnnGetRNNParamsSize! Why not use that?

src/rnn.jl
```diff
@@ -635,7 +659,7 @@ function rnntest(r::RNN, ws, x, hx=nothing, cx=nothing;
                  hy = (hx != nothing),
                  cy = (cx != nothing && r.mode == 2),
                  o...)
-    if r.direction == 1; error("rnntest bidirectional not implemented yet"); end
+    #if r.direction == 1; error("rnntest bidirectional not implemented yet"); end
     if r.dropout != 0; error("rnntest dropout not implemented yet"); end
```
denizyuret: I think we can add support for dropout fairly easily.  Just have to make sure exactly where it is applied.  Note that it has no effect in a single layer network according to cudnn docs.

src/rnn.jl
```diff
@@ -675,52 +738,84 @@ function rnntest(r::RNN, ws, x, hx=nothing, cx=nothing;
...
+                ht = tanh.(Whxt .+ rt .* (Rh' * hs[i] .+ bRh) .+ bWh)
+                hs[i] = (1 .- it) .* ht .+ it .* hs[i]
             end
-            push!(ys, hs[L])
         end
     else
         error("RNN not supported")
     end
-    y = reshape(hcat(ys...), ysize)
+    y = r.direction == 0 ? reshape(hcat(ys...), ysize) : ys
```
denizyuret: You don't put the output in the right order?  We want our output to match cudnn exactly...

src/rnn.jl
```diff
@@ -675,52 +738,84 @@ function rnntest(r::RNN, ws, x, hx=nothing, cx=nothing;
...
     hyout = hy ? reshape(hcat(hs...), hsize) : nothing
     cyout = cy && r.mode == 2 ? reshape(hcat(cs...), hsize) : nothing
     return (y,hyout,cyout,nothing)
 end
```
denizyuret: This whole thing can be tested by just making a small change in test/rnn.jl, just edit:
```
    for M=(:relu,:tanh,:lstm,:gru), L=1:2, I=(:false,:true)
```
to
```
    for M=(:relu,:tanh,:lstm,:gru), L=1:2, I=(:false,:true), BI=(:false,:true)
```
and use `bidirectional=BI` in the `rnninit` commands.

denizyuret:
Great work!  I added some minor comments.  Testing with the current
test/rnn.jl should be fairly easy, see the last comment.  Ilker, once the
cpu implementation passes the gpu comparison tests we can pull in your
demos and also send a pull request to dynet people.

src/rnn.jl
denizyuret: OK, I made this the default behavior of cudnnhandle in the latest master.

src/rnn.jl
denizyuret: I looked at this and it is difficult to do it in general.  The user will get confused about when view is supported and when it is not with KnetArrays.  However for this particular case, instead of defining a new function you can safely say:
```
Base.view(a::KnetArray,i::AbstractUnitRange)=a[i]
```

cangumeli:
Well, first let me explain what I understood from multi-layer bidirectional RNN to answer some of the questions:
1. Each layer's input is the previous layer's output (not the hidden state). This whole output array is processed in reverse directions by the next layer.

2. A layer's output is vertical concatenation of forward direction layer's state history and reverse of backward direction layer's state history. For example, the first time step of the output is vertical concatenation of first hidden state of the forward rnn and last hidden state of the reverse rnn, both correspond to first time step of the input. Because of this behaviour:
a) Input matrices starting from the second bidirectional layer (or third layer) have sizes `(2H, H)`.
b) All time-steps of one layer must be finished before the following layer's execution started (that's why order of for-loops had to be different).
c) The last `X` produced by the `bidirect` function exactly matches the `y` produced by cuDNN.

3. Hidden states are returned as if reverse directions are layers, which was already implemented when I started.

cangumeli:
On the `view` issue;

Supporting views in `KnetArray` is easy, it is `getindex`. The problem is I think we don't support`view` properly in CPU. In CPU, `view` produces things like `SubArray` and `ReshapedArray`. I don't know how these arrays work and how they can be used. Also, `view`  was not an AutoGrad primitive last time I checked, while `rnnparams` must be differentiable for `rnntest` to work when training.

src/rnn.jl
cangumeli:
```
julia> r,w = rnninit(5,5; numLayers=2, rnnType=:relu, bidirectional=true);

julia> for mat in rnnparams(r,w)
           println(size(mat))
       end
(5, 5)
(5, 5)
(5, 5)
(5, 5)
(10, 5)
(5, 5)
(10, 5)
(5, 5)
(5,)
(5,)
(5,)
(5,)
(5,)
(5,)
(5,)
(5,)
```

src/rnn.jl
denizyuret: In light of your latest comments, we could just get rid of views all together.  Did you try them with AutoGrad, and did they fail?

denizyuret:
OK, if y, hy, cy match then I am happy.  With view, if it doesn't work for AutoGrad or it doesn't improve the performance by much, we can make it false by default and get back to it later.  Did test/rnn.jl work?

src/rnn.jl
cangumeli: I tried them 2 weeks ago while working on `batchnorm`. It was failed in gradchecks, so I did not push it. I don't remember the actual error log, however. Also, typing is a problem
```
julia> a = rand(10,10);

julia> b = view(a, 1:2,1:3)
2×3 SubArray{Float64,2,Array{Float64,2},Tuple{UnitRange{Int64},UnitRange{Int64}},false}:
 0.507469  0.979536  0.292401
 0.211171  0.509129  0.335866

julia> isa(b, Array{Float64})
false

julia> isa(a, Array{Float64})
true
```

cangumeli:
> OK, if y, hy, cy match then I am happy. With view, if it doesn't work for AutoGrad or it doesn't improve the performance by much, we can make it false by default and get back to it later. Did test/rnn.jl work?

test/rnn.jl works. `useview` is `false` by default.

PR #229, opened 2017-12-15
[WIP] more general indexing for KnetArray

CarloLucibello:
Add general `getindex` and `setindex!` fallbacks to deal with cases like these, currently giving errors:
```julia
julia> a=KnetArray(rand(3,3,2));

julia> a[1:2,:,1]
ERROR: MethodError: no method matching getindex(::Knet.KnetArray{Float64,3}, ::UnitRange{Int64}, ::Colon, ::Int64)
Stacktrace:
 [1] macro expansion at ./REPL.jl:97 [inlined]
 [2] (::Base.REPL.##1#2{Base.REPL.REPLBackend})() at ./event.jl:73

julia> a[1:2,1:2,1]
ERROR: MethodError: no method matching getindex(::Knet.KnetArray{Float64,3}, ::UnitRange{Int64}, ::UnitRange{Int64}, ::Int64)
Stacktrace:
 [1] macro expansion at ./REPL.jl:97 [inlined]
 [2] (::Base.REPL.##1#2{Base.REPL.REPLBackend})() at ./event.jl:73
```
It uses some machinery from Base to fallback to indexing with a vector of linear indexes. 
Most credits go to @carlobaldassi

Still, have to test properly and to drop scalar dimensions in reshape.


denizyuret:
Let's performance check these implementations before releasing them.  At least let's make sure that they are faster than a copy-to-cpu, index, copy-to-gpu round-trip solution.

CarloLucibello:
> Let's performance check these implementations before releasing them. At least let's make sure that they are faster than a copy-to-cpu, index, copy-to-gpu round-trip solution.

will do

PR #234, opened 2017-12-20
include batchnorm docs in function reference

ilkerkesen:
Include batchnorm documentation in function reference. Fixed one typo also.


denizyuret:
You can check to see if the latest Knet docs reflect your changes.

Issue #212, opened 2017-12-05
change logp  to log_softmax


CarloLucibello:
Not really an issue but merely a suggestion. I find the name `logp`, which is supposed to stand for log probabilities, to be quiet unintuitive, and also too close to `log1p`. Could we rename it to `log_softmax`, or export `log_softmax` as an alias?

Also, it would be nice to also have the `softmax` function exported, possibily using the cuda backend.

Bye
C

PR #237, opened 2017-12-21
fix logp for karray

CarloLucibello:
Previous version was focused on matrices, now it is generally improved both for matrices and tensors. 
This is done exploiting the `mode=1` (cumulating only on the channel dimension) of the cuda softmax call.

```julia
# PRE

julia> k = KnetArray(rand(10,100)) # 2d tensor

julia> @btime logp($k,1);
  4.111 μs (23 allocations: 1.08 KiB)

julia> @btime logp($k,2);
  9.929 μs (39 allocations: 1.42 KiB)


julia> k = KnetArray(rand(Float32,10,10,100)); # 3d tensor 

julia> @btime logp($k,1);
  4.195 μs (21 allocations: 1.00 KiB)

julia> @btime logp($k,2);
ERROR: Transpose is supported only for 2D KnetArrays
Stacktrace:
 [1] transpose(::Knet.KnetArray{Float32,3}) at /home/carlo/Git/Knet.jl/src/linalg.jl:89
 [2] logp(::Knet.KnetArray{Float32,3}, ::Int64, ::Vararg{Int64,N} where N) at /home/carlo/Git/Knet.jl/src/loss.jl:19
 [3] ##core#1420(::Knet.KnetArray{Float32,3}) at /home/carlo/.julia/v0.6/BenchmarkTools/src/execution.jl:312
 [4] ##sample#1421(::BenchmarkTools.Parameters) at /home/carlo/.julia/v0.6/BenchmarkTools/src/execution.jl:318
 [5] #_run#7(::Bool, ::String, ::Array{Any,1}, ::Function, ::BenchmarkTools.Benchmark{Symbol("##benchmark#1419")}, ::BenchmarkTools.Parameters) at /home/carlo/.julia/v0.6/BenchmarkTools/src/execution.jl:346
 [6] (::BenchmarkTools.#kw##_run)(::Array{Any,1}, ::BenchmarkTools.#_run, ::BenchmarkTools.Benchmark{Symbol("##benchmark#1419")}, ::BenchmarkTools.Parameters) at ./<missing>:0
 [7] anonymous at ./<missing>:?
 [8] #run_result#19(::Array{Any,1}, ::Function, ::BenchmarkTools.Benchmark{Symbol("##benchmark#1419")}, ::BenchmarkTools.Parameters) at /home/carlo/.julia/v0.6/BenchmarkTools/src/execution.jl:40
 [9] (::BenchmarkTools.#kw##run_result)(::Array{Any,1}, ::BenchmarkTools.#run_result, ::BenchmarkTools.Benchmark{Symbol("##benchmark#1419")}, ::BenchmarkTools.Parameters) at ./<missing>:0
 [10] #run#21(::Array{Any,1}, ::Function, ::BenchmarkTools.Benchmark{Symbol("##benchmark#1419")}, ::BenchmarkTools.Parameters) at /home/carlo/.julia/v0.6/BenchmarkTools/src/execution.jl:63
 [11] (::Base.#kw##run)(::Array{Any,1}, ::Base.#run, ::BenchmarkTools.Benchmark{Symbol("##benchmark#1419")}, ::BenchmarkTools.Parameters) at ./<missing>:0
 [12] warmup(::BenchmarkTools.Benchmark{Symbol("##benchmark#1419")}) at /home/carlo/.julia/v0.6/BenchmarkTools/src/execution.jl:96
 [13] macro expansion at ./REPL.jl:97 [inlined]
 [14] (::Base.REPL.##1#2{Base.REPL.REPLBackend})() at ./event.jl:73

julia> @btime logp($k,3);
  20.335 μs (78 allocations: 2.69 KiB)
```

```julia
# AFTER

julia> k = KnetArray(rand(10,100)) # 2d tensor

julia> @btime logp($k,1);
  3.997 μs (24 allocations: 1.13 KiB)

julia> @btime logp($k,2);
  3.987 μs (24 allocations: 1.13 KiB)

julia> k = KnetArray(rand(Float32,10,10,100)); # 3d tensor 

julia> @btime logp($k,1);
  4.018 μs (25 allocations: 1.17 KiB)

julia> @btime logp($k,2);
  4.037 μs (25 allocations: 1.16 KiB)

julia> @btime logp($k,3);
  4.013 μs (24 allocations: 1.16 KiB)
```

### DIFF
--- a/prof/softmax.jl
+++ b/prof/softmax.jl
@@ -1,5 +1,5 @@
 using Knet, AutoGrad
-using Knet: @cuda, cudnnhandle, Cptr, TD, cudnnSoftmaxForward, cudnnSoftmaxBackward, TD4, _logp
+using Knet: @cuda, cudnnhandle, Cptr, TD, cudnnSoftmaxForward, cudnnSoftmaxBackward, _logp
 using BenchmarkTools
 
 # algo=2 corresponds to logp(x,1)
--- a/prof/softmax.jl
+++ b/prof/softmax.jl
@@ -52,16 +52,13 @@ back2(x,d...)=(for i=1:rep(x); grad2(x,d...); end; cds())
 
 # dimensions: 0/1/2, forw/back, model0/1, x1:5
 
-for d in (0,1,2)
+for d in ((1,2),1,2)
     for f in (logp0, logp1, logp2, back0, back1, back2)
-        @printf("%s(x,%d)",f,d)
+        print("$f(x, $d)")
         for x in (x1,x2,x3,x4,x5)
             xt = x.'
             knetgc()
-            b = (d==0 ? (@benchmark $f($x)) :
-                 d==1 ? (@benchmark $f($x,1)) :
-                 d==2 ? (@benchmark $f($xt,2)) :
-                 error())
+            b = @benchmark $f($x, $d))
             times = b.times ./ rep(x)
             # @printf("\t%dx%d/%.1e/%.1e/%.1e", length(times), rep(x), minimum(times), median(times), mean(times))
             @printf("\t%.1e", minimum(times))
--- a/src/Knet.jl
+++ b/src/Knet.jl
@@ -28,7 +28,7 @@ include("linalg.jl");           export mat # matmul, axpy!, transpose, (i)permut
 include("conv.jl");             export conv4, pool, deconv4, unpool
 include("batchnorm.jl");        export batchnorm, bnmoments, bnparams
 include("rnn.jl");              export rnnforw, rnninit, rnnparam, rnnparams
-include("loss.jl");             export logp, logsumexp, nll, accuracy
+include("loss.jl");             export logp, logsumexp, nll, accuracy, softmax, logsoftmax
 include("dropout.jl");          export dropout
 include("update.jl"); 		export Sgd, Momentum, Nesterov, Adam, Adagrad, Adadelta, Rmsprop, update!, optimizers
 include("distributions.jl"); 	export gaussian, xavier, bilinear
--- a/src/loss.jl
+++ b/src/loss.jl
@@ -1,3 +1,5 @@
+const DimsOrVec = Union{Dims, Vector}
+
 """
 
     logp(x,[dims])
@@ -10,12 +12,23 @@ over the whole `x`, otherwise the normalization is performed over the
 given dimensions.  In particular, if `x` is a matrix, `dims=1`
 normalizes columns of `x` and `dims=2` normalizes rows of `x`.
 
+Calls to `logsoftmax(x)` is `logp(x, 1)`, and `softmax(x)` is 
+equivalent to `exp.(log(x,1)`. 
 """
 logp(x) = logp(x, tuple(1:ndims(x)...)) 
+logp(x, dims::DimsOrVec) = _logp(x, dims) # generic fallback
 logp(x, d::Int) = logp(x, (d,))
-logp(x, dims::Dims) = _logp(x, dims) # generic fallback
 
-function logp(x::A, dims::Dims) where A <: Union{KnetArray, Rec{KnetArray}}
+# specialized version avoid a little overhead of logp(x::A, dims)
+function logp(x::A, d::Int) where A <: Union{KnetArray, Rec{KnetArray}}
+    @assert 1 <= d <= ndims(x)  
+    sz = size(x)
+    x = reshape(x, (prod(sz[1:d-1]), 1, sz[d], prod(sz[d+1:end])))
+    x = cudnnSoftmaxForward(x, mode=1, algo=2)
+    reshape(x, sz)
+end
+
+function logp(x::A, dims::DimsOrVec) where A <: Union{KnetArray, Rec{KnetArray}}
     d = sort(union(dims))
     @assert all(1 <= d <= ndims(x) for d in d)  
     if length(d) == 0
@@ -24,7 +37,7 @@ function logp(x::A, dims::Dims) where A <: Union{KnetArray, Rec{KnetArray}}
             reshape(cudnnSoftmaxForward(reshape(x,(1,1,n,1)),algo=2),size(x)))
     elseif areconsecutives(d)
         sz = size(x)
-        x = reshape(x, (prod(sz[1:d[1]-1]), 1, prod(sz[d]), prod(sz[d[end]+1:end])))
+        @inbounds  x = reshape(x, (prod(sz[1:d[1]-1]), 1, prod(sz[d]), prod(sz[d[end]+1:end])))
         x = cudnnSoftmaxForward(x, mode=1, algo=2)
         reshape(x, sz)
     else
@@ -34,8 +47,21 @@ end
 
 areconsecutives(d) = all(d[i+1] == d[i]+1 for i=1:length(d)-1)
 
-logsoftmax = logp
-softmax(x, d...) = exp.(logsoftmax(x, d...)) 
+"""
+    logsoftmax(x)
+
+Equivalent to `logp(x, 1)`. See also `sotfmax`. 
+"""
+logsoftmax(x) = logp(x, 1)
+
+"""
+    softmax(x)
+
+The softmax function typically used in classification.
+It is equivalen to `exp.(logp(x, 1))`. 
+See also `logsoftmax`.
+"""
+softmax(x) = exp.(logsoftmax(x)) 
 
 # Math for the cross-entropy loss: x is unnormalized input, p is
 # target probabilities, q is estimated probabilities. Read left column
@@ -57,7 +83,7 @@ softmax(x, d...) = exp.(logsoftmax(x, d...))
 _logp(x) = _logp(x, tuple(1:ndims(x)...))
 _logp(x, d::Int) = _logp(x, (d,))
 
-function _logp(x, dims::Dims)
+function _logp(x, dims::DimsOrVec)
     @assert all(1 <= d <= ndims(x) for d in dims)  
     xval = getval(x)
     if isa(xval,Number)
@@ -78,7 +104,7 @@ function _logp(x, dims::Dims)
     end
 end
 
-function _logpback(x,y,dy,dims::Dims)
+function _logpback(x,y,dy,dims::DimsOrVec)
     xval = getval(x)
     if isa(xval,Number)
         return zero(xval)
@@ -96,7 +122,7 @@ function _logpback(x,y,dy,dims::Dims)
 end
 
 # dy should be -p and y=logq so this should give us -p+q
-@primitive  _logp(x,dims::Dims),dy,y  _logpback(x,y,dy,dims)
+@primitive  _logp(x,dims::DimsOrVec),dy,y  _logpback(x,y,dy,dims)
 
 #=
 typedef enum
--- a/src/loss.jl
+++ b/src/loss.jl
@@ -33,6 +33,9 @@ function logp(x::A, d::Integer) where A <: Union{KnetArray, Rec{KnetArray}}
     x = reshape(x, sz)
 end
 
+logsoftmax = logp
+softmax(x, d...) = exp.(logsoftmax(x, d...)) 
+
 # Math for the cross-entropy loss: x is unnormalized input, p is
 # target probabilities, q is estimated probabilities. Read left column
 # down, right column (loss gradients) back up.
--- a/src/loss.jl
+++ b/src/loss.jl
@@ -13,11 +13,7 @@ normalizes columns of `x` and `dims=2` normalizes rows of `x`.
 """
 function logp(x,d...)
     if isa(getval(x), KnetArray)
-        if d==(1,)
-            cudnnSoftmaxForward(x,algo=2)
-        elseif d==(2,)
-            cudnnSoftmaxForward(x.',algo=2).'
-        elseif d==()
+        if d==()
             n = length(x)
             (n > 20000 ? _logp(x) : # see Knet/prof/softmax.jl for timing info
              reshape(cudnnSoftmaxForward(reshape(x,(1,1,n,1)),algo=2),size(x)))
@@ -27,6 +23,14 @@ function logp(x,d...)
     else
         _logp(x,d...) # fall back on old implementation if not KnetArray
     end
+end
+
+function logp(x::A, d::Integer) where A <: Union{KnetArray, Rec{KnetArray}}
+    @assert 1 <= d <= ndims(x)  
+    sz = size(x)
+    x = reshape(x, (prod(sz[1:d-1]), 1, sz[d], prod(sz[d+1:end])))
+    x = cudnnSoftmaxForward(x, mode=1, algo=2)
+    x = reshape(x, sz)
 end
 
 # Math for the cross-entropy loss: x is unnormalized input, p is
--- a/src/loss.jl
+++ b/src/loss.jl
@@ -1,43 +1,25 @@
-const DimsOrVec = Union{Dims, Vector}
-
 """
-
-    logp(x,[dims])
+    logp(x, dims=1)
 
 Treat entries in `x` as as unnormalized log probabilities and return
 normalized log probabilities.
 
 `dims` is an optional argument, if not specified the normalization is
-over the whole `x`, otherwise the normalization is performed over the
+over the first dimension of `x`, otherwise the normalization is performed over the
 given dimensions.  In particular, if `x` is a matrix, `dims=1`
-normalizes columns of `x` and `dims=2` normalizes rows of `x`.
+normalizes columns of `x`, `dims=2` normalizes rows of `x` and
+`dims=(1,2)` normalizes the whole matrix.  
 
-Calls to `logsoftmax(x)` is `logp(x, 1)`, and `softmax(x)` is 
-equivalent to `exp.(log(x,1)`. 
+Calls to `logsoftmax` are equivalent to `logp`, and `softmax(x,dims)` is 
+equivalent to `exp.(logp(x,dims)`. 
 """
-logp(x) = logp(x, tuple(1:ndims(x)...)) 
-logp(x, dims::DimsOrVec) = _logp(x, dims) # generic fallback
-logp(x, d::Int) = logp(x, (d,))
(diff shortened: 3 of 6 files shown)


CarloLucibello:
since I was at it I added aliases for `softmax` and `logsoftmax` (ref #212 )

denizyuret:
Looks great!  What is the semantics of logp(x,[1,2,3]) and is it consistent with sum(x,[1,2,3]) etc?

CarloLucibello:
`logp` is different from `sum`, since the output of `log(x,dims)` has the same size of `x`. In any case, I didn't look into `logp`  over multiple dimensions yet, but it should be quite easy to resort to cudaSoftmax when `dims` contains consecutive indices, and fall back to `_logp(x,dims)` otherwise.

denizyuret:
Yes they are different but the interpretation of the dims argument is
intended to be similar: i.e whatever sum() sums over, logp normalizes over
with the same dims argument.

CarloLucibello:
I extended the cuda softmax call to multidimensional regions with consecutive indexes. The whole interface is consistent with `sum(x, dims)` Base interface, has these examples show:

```julia
julia> x=KnetArray(rand(Float32,2,3,4))
2×3×4 Knet.KnetArray{Float32,3}:
[:, :, 1] =
 0.214499  0.827421   0.154492
 0.468541  0.0453421  0.571895

[:, :, 2] =
 0.621309  0.853617  0.511255
 0.7107    0.766149  0.541774

[:, :, 3] =
 0.407307  0.961725  0.0806607
 0.295493  0.856248  0.279023

[:, :, 4] =
 0.749372  0.37773  0.603461
 0.621515  0.82864  0.110373

julia> d=1; sum(softmax(x,d), d)
1×3×4 Knet.KnetArray{Float32,3}:
[:, :, 1] =
 1.0  1.0  1.0

[:, :, 2] =
 1.0  1.0  1.0

[:, :, 3] =
 1.0  1.0  1.0

[:, :, 4] =
 1.0  1.0  1.0
julia> d=2; sum(softmax(x,d), d)
2×1×4 Knet.KnetArray{Float32,3}:
[:, :, 1] =
 1.0
 1.0

[:, :, 2] =
 1.0
 1.0

[:, :, 3] =
 1.0
 1.0

[:, :, 4] =
 1.0
 1.0

julia> d=3; sum(softmax(x,d), d)
2×3×1 Knet.KnetArray{Float32,3}:
[:, :, 1] =
 1.0  1.0  1.0
 1.0  1.0  1.0

julia> d=(1,2); sum(softmax(x,d), d)
1×1×4 Knet.KnetArray{Float32,3}:
[:, :, 1] =
 1.0

[:, :, 2] =
 1.0

[:, :, 3] =
 1.0

[:, :, 4] =
 1.0

julia> d=(1,3); sum(softmax(x,d), d)
1×3×1 Knet.KnetArray{Float32,3}:
[:, :, 1] =
 1.0  1.0  1.0
```
With respect to the first commit of this PR I'm seeing a performance degradation I should investigate:
```julia
julia> @btime logp($k,1);
  5.553 μs (45 allocations: 2.33 KiB)

julia> @btime logp($k,2);
  5.377 μs (45 allocations: 2.31 KiB)

julia> @btime logp($k,3);
  5.522 μs (44 allocations: 2.31 KiB)
```

CarloLucibello:
The performance lost  4 μs -> 5.55 μs  in the new version is due to to the sort(union(dims)), to the assert and to some extra indexing in the reshape. Of course this will be less relevant for larger array sizes, but if @denizyuret thinks it is worth it I can add the extra method
```julia
function logp(x::A, d::Integer) where A <: Union{KnetArray, Rec{KnetArray}}
           @assert 1 <= d <= ndims(x)  
           sz = size(x)
           x = reshape(x, (prod(sz[1:d-1]), 1, sz[d], prod(sz[d+1:end])))
           x = cudnnSoftmaxForward(x, mode=1, algo=2)
           x = reshape(x, sz)
       end
```

CarloLucibello:
I added the `logp(x::A, d::Integer)`, and relaxed some signatures. Know we have `logp(x, 1)`, `logp(x, (1,2))` and `logp(x, [1,2])` as with `sum`, `prod`, `maximum`, ... in Base.

Also now `logsoftmax(x)=logp(x,1)`  and `softmax(x) = exp.(logp(x,1))`.

It should be ready to go if tests pass

denizyuret:
Thanks, will review and merge.

denizyuret:
About the alternatives:
```
logsoftmax(x)=logp(x,1)
```
I think we should make logsoftmax a proper synonym to avoid confusion, i.e. const logsoftmax = logp.  People may want to use the dims argument with logsoftmax as well (e.g. using row major arrays for performance reasons).  

```
softmax(x) = exp.(logp(x,1)).
```
There is a mode in cudnn that directly computes this. I will check if we are using it. Also, similar to logsoftmax, why not define softmax(x,d) with a regular dims argument?

We should test things with:
```
isapprox(logp(x,d), log.(exp.(x) ./ sum(exp.(x), d)))
isapprox(softmax(x,d), exp.(x) ./ sum(exp.(x), d)))
```

prof/softmax.jl
```diff
@@ -1,5 +1,5 @@
 using Knet, AutoGrad
-using Knet: @cuda, cudnnhandle, Cptr, TD, cudnnSoftmaxForward, cudnnSoftmaxBackward, TD4, _logp
+using Knet: @cuda, cudnnhandle, Cptr, TD, cudnnSoftmaxForward, cudnnSoftmaxBackward, _logp
```
denizyuret: Never used TD4 here, why import?

src/Knet.jl
```diff
@@ -28,7 +28,7 @@ include("linalg.jl");           export mat # matmul, axpy!, transpose, (i)permut
 include("conv.jl");             export conv4, pool, deconv4, unpool
 include("batchnorm.jl");        export batchnorm, bnmoments, bnparams
 include("rnn.jl");              export rnnforw, rnninit, rnnparam, rnnparams
-include("loss.jl");             export logp, logsumexp, nll, accuracy
+include("loss.jl");             export logp, logsumexp, nll, accuracy, softmax, logsoftmax
```
denizyuret: See comments in PR for softmax and logsoftmax interface.

src/loss.jl
```diff
@@ -106,7 +145,7 @@ function cudnnSoftmaxForward{T}(x::KnetArray{T}; algo=0, mode=0, alpha=1, handle
     y = similar(x)
     @cuda(cudnn, cudnnSoftmaxForward,
           (Cptr, Cint, Cint, Ptr{T}, Cptr, Ptr{T}, Ptr{T}, Cptr, Ptr{T}),
-          handle, algo, mode, Ref(T(alpha)), TD4(x), x, Ref(T(beta)), TD4(y), y)
+          handle, algo, mode, Ref(T(alpha)), TD(x), x, Ref(T(beta)), TD(y), y)
```
denizyuret: Why did we go back to TD from TD4 here?  cudnn does not like 1-2 dimensional arrays, TD4 reshapes to make cudnn happy.

src/reduction.jl
```diff
@@ -2,14 +2,15 @@
...
 # ai is the accumulator, xi is the array element
-if VERSION >= v"0.6.0"
-    import AutoGrad: sumabs_, sumabs2_, minabs_, maxabs_
-    Base.sum(::typeof(abs), x::KnetArray, d...) = sumabs_(x,d...);
-    Base.sum(::typeof(abs2), x::KnetArray, d...) = sumabs2_(x,d...);
-    Base.maximum(::typeof(abs), x::KnetArray, d...) = maxabs_(x,d...);
-    Base.minimum(::typeof(abs), x::KnetArray, d...) = minabs_(x,d...);
-    reduced_dims_compat(dims,region)=map(last, Base.reduced_indices(map(Base.OneTo, dims), region))
-    reduction_ops = [
+import AutoGrad: sumabs_, sumabs2_, minabs_, maxabs_
```
denizyuret: Any reason things changed here?  Did you just clear the pre-0.6 pieces?  Let's try to keep the changes to a relevant minimum so it is easy to review/test...

src/reduction.jl
```diff
@@ -49,8 +36,12 @@ function reduction_op(f, j=f, o...)
                 y=ccall(($F20,$libknet8),$T,(Cint,Ptr{$T}),length(x),x) # do not use @knet8, return not Void
                 @gs; return y
             end
+            
+            $J(x::KnetArray{$T}, d::Int) = $J(x, (d,))
```
denizyuret: Oh no, I see you are standardizing the dims argument here.  Was there a problem with the original implementation?  Or did you decide to change following to convention in Base?

test/broadcast.jl
```diff
@@ -68,12 +63,12 @@ srand(42)
                     @dbg f,t,n1,n2
                     a1 = rand11(f,t,n1)
                     a2 = rand11(f,t,n2)+t(1)
-                    @test gradcheck(f1, Any[a1, a2])
+                    @test gradcheck(f1, [a1, a2])
```
denizyuret: We sometimes force Any[] type so when AutoGrad tries to replace elements of an array with their ::Rec counterparts we don't get an error, because the outer array type only allows for numeric arrays and not Rec's etc.  This was not the case here?

test/loss.jl
```diff
@@ -15,6 +17,28 @@ include("header.jl")
...
+            @test gradcheck(f,k)
+            @test isapprox(f(a),f(k))
+            for d in [1,2,3]
+                @test gradcheck(f,k,d)
+                @test isapprox(f(a,d),f(k,d))
+            end
+            for dims in [(1,2), (1,3), (3,1)]
+                @test isapprox(f(a,dims),f(k,dims))
+            end
+        end
```
denizyuret: I guess we can add somewhere here the definitional tests comparing logp, softmax etc with their exp.(x) / sum(exp.(x), dims) counterparts.

denizyuret:
Having thought about this some more I can see the rationale of having logsoftmax and softmax be defined with dims=1 as default, making them similar to other frameworks.  On the other hand I also like the generic Julia way of defining these for any dims argument, so the user has the ability to normalize across any dimensions they want. So I am torn.

prof/softmax.jl
CarloLucibello: don't know, I gues some historical residue

src/loss.jl
CarloLucibello: I removed `TD4` because I'm reshaping the inputs before calling `cudnnSoftmaxForward`

src/reduction.jl
CarloLucibello: Just clear pre-0.6, I cannot resist removing all those version check when I see them, sorry

src/reduction.jl
CarloLucibello: I cannot remember why I add this. The main drive was to be coherent with Base. I'll look into it

test/loss.jl
CarloLucibello: will do

CarloLucibello:
I think that `dims=1` is a much saner default for `softmax` and `logsoftmax`. For other cases we could define `softmax(x, dims)` and have user spare some keystrokes with respect to `exp.(logp(x,dims)`

denizyuret:
In that case we should define all 3 functions with dims=1 default to avoid confusion.  The interpretation of dims arg will be similar to sum but the default will be different.  In that case logsoftmax can just be an alias for logp.  softmax needs a proper cudnn call without the extra exp operation (see cudnn docs or my comments in code).  Need to use algo=0 or 1 for that.  1 is a bit slower but safer.  We can benchmark and decide the default.  We can also have a keyword arg for softmax so the user can override the default.
```
typedef enum                                                                                                               
{                                                                                                                          
    CUDNN_SOFTMAX_FAST     = 0,         /* straightforward implementation */                                               
    CUDNN_SOFTMAX_ACCURATE = 1,         /* subtract max from every point to avoid overflow */                              
    CUDNN_SOFTMAX_LOG      = 2                                                                                             
} cudnnSoftmaxAlgorithm_t;                                                                                                 
```

CarloLucibello:
> In that case we should define all 3 functions with dims=1 default to avoid confusion. The interpretation of dims arg will be similar to sum but the default will be different. In that case logsoftmax can just be an alias for logp. softmax needs a proper cudnn call without the extra exp operation (see cudnn docs or my comments in code). Need to use algo=0 or 1 for that. 1 is a bit slower but safer. We can benchmark and decide the default. We can also have a keyword arg for softmax so the user can override the default.

Done. Here I kept `softmax` and `logsoftmax` as separate functions, another option would be to merge them together and have `algo=1,2,3`. I'd prefer to have them separate. 

Also, I deleted this branch 
```julia
 n = length(x)
 (n > 20000 ? _logp(x) : # see Knet/prof/softmax.jl for timing info
                    reshape(cudnnSoftmaxForward(reshape(x,(1,1,n,1)),algo=2),size(x)))
```
since I could not reproduce the supposed better performance of `_logp` over the cuda call for large `n`:
```julia
julia> k=KnetArray(rand(Float32,1000,100));

julia> @btime Knet._logp(k, (1,2));
  63.823 μs (62 allocations: 1.83 KiB)

julia> @btime logp(k, (1,2));
  5.258 μs (37 allocations: 2.05 KiB)
```

denizyuret:
Are you turning on gpu sync when you are testing?  If not, the gpu calls
are asynchronous and they would fool our @btime into thinking no time has
been spent at all.  Take a look at prof/softmax.jl -- not syncing is not
good, syncing at every call is too costly, so I write functions that sync
every n calls where n is determined based on array size.

CarloLucibello:
I did some proper benchmarking as you suggested, essentially using prof/softmax.jl, and it appears that generally mode=1 of cudaSoftmax has worse performance than other methods:
```
# batch 100
*logp0 = _logp # pure julia
*logp1 = logp_  # cuda old
*logp2 = logp # cuda new

x1 = ka(randn(Float32,10,100))
x2 = ka(randn(Float32,100,100))
x3 = ka(randn(Float32,1000,100))
x4 = ka(randn(Float32,10000,100))
x5 = ka(randn(Float32,100000,100))

logp0(x, (1, 2))        4.1e+04 4.0e+04 `5.1e+04 2.2e+05 1.8e+06`
logp1(x, (1, 2))        `3.9e+03 8.4e+03` 1.1e+05 1.8e+06 1.8e+07
logp2(x, (1, 2))        5.8e+03 1.6e+04 1.1e+05 1.8e+06 1.8e+07

back0(x, (1, 2))        1.2e+05 1.2e+05 `1.4e+05 4.7e+05 3.7e+06`
back1(x, (1, 2))        `7.4e+04 7.4e+04` 2.6e+05 3.3e+06 3.3e+07
back2(x, (1, 2))        7.8e+04 7.9e+04 2.6e+05 3.3e+06 3.3e+07

logp0(x, 1)             1.9e+04 1.9e+04 2.1e+04 1.9e+05 1.7e+06
logp1(x, 1)             `4.5e+03 4.6e+03 6.0e+03 6.9e+04 7.8e+05`
logp2(x, 1)             5.5e+03 5.8e+03 8.8e+03 8.4e+04 7.8e+05

back0(x, 1)             8.9e+04 8.9e+04 8.9e+04 4.1e+05 3.6e+06
back1(x, 1)             `7.4e+04 7.5e+04 7.4e+04 2.0e+05 1.8e+06`
back2(x, 1)             1.7e+05 1.7e+05 3.5e+05 1.9e+06 1.6e+07

logp0(x, 2)             2.0e+04 2.0e+04 `2.9e+04` 2.9e+05 3.8e+06
logp1(x, 2)             `1.0e+04 1.2e+04` 3.5e+04 3.5e+05 3.2e+06
logp2(x, 2)             3.7e+04 3.7e+04 3.8e+04 `1.3e+05 1.0e+06`

back0(x, 2)             `9.1e+04 9.0e+04 9.8e+04` 5.7e+05 6.7e+06
back1(x, 2)             1.0e+05 1.0e+05 1.2e+05 6.8e+05 6.1e+06
back2(x, 2)             1.1e+05 1.1e+05 1.2e+05 `2.6e+05 2.0e+06`


# batch 1000
x1 = ka(randn(Float32,10,1000))
x2 = ka(randn(Float32,100,1000))
x3 = ka(randn(Float32,1000,1000))
x4 = ka(randn(Float32,10000,1000))

logp0(x, (1, 2))        4.1e+04 `5.7e+04 2.2e+05 1.8e+06`
logp1(x, (1, 2))        `8.6e+03` 1.1e+05 1.8e+06 1.8e+07
logp2(x, (1, 2))        1.6e+04 1.1e+05 1.8e+06 1.8e+07

back0(x, (1, 2))        1.2e+05 `1.4e+05 4.7e+05 3.7e+06`
back1(x, (1, 2))        `7.5e+04` 2.6e+05 3.4e+06 3.3e+07
back2(x, (1, 2))        1.2e+05 `1.4e+05 4.7e+05 3.7e+06`

logp0(x, 1)             2.2e+04 3.0e+04 1.9e+05 1.7e+06
logp1(x, 1)             `3.9e+03 6.0e+03 4.6e+04 4.0e+05`
logp2(x, 1)             3.3e+04 3.5e+04 7.3e+04 7.8e+05

back0(x, 1)             8.9e+04 9.7e+04 4.2e+05 3.5e+06
back1(x, 1)             7.5e+04 7.6e+04 1.7e+05 1.3e+06
back2(x, 1)             9.0e+04 9.7e+04 4.2e+05 3.5e+06

logp0(x, 2)             `2.0e+04 2.9e+04` 2.5e+05 2.6e+06
logp1(x, 2)             3.2e+04 3.5e+04 `1.3e+05` 1.4e+06
logp2(x, 2)             3.5e+05 3.5e+05 6.4e+05 `1.2e+06`

back0(x, 2)             9.2e+04 9.9e+04 5.3e+05 5.2e+06
back1(x, 2)             1.1e+05 1.1e+05 3.0e+05 2.7e+06
```
In the end I took the conservative approach and reverted most work done in this PR. The contribution of this PR now is
- fix error when computing `logp(x, 2)` on non bidimensional `x`
-  conform `logp` api to that of reduction methods in Base, except for when the `dims` argument is omitted (defaults to 1)
- export the alias `logsoftmax` for `logp`
- define and export `softmax` for julia and knetarray

CarloLucibello:
closing in favor of #313

PR #240, opened 2017-12-25
fix mnist data dependency, add Images package

ilkerkesen:
I did forget to add Images as dependency. Image resizing used in mnistgrid procedure, that's why it needs Images.

### DIFF
--- a/data/mnist.jl
+++ b/data/mnist.jl
@@ -1,4 +1,4 @@
-for p in ("GZip",)
+for p in ("GZip","Images")
     Pkg.installed(p) == nothing && Pkg.add(p)
 end
 


denizyuret:
Should we generalize this as an image grid and add it to data/imagenet.jl ?  Advantages:
1. Loading mnist will be faster for people not interested in visualization (imagenet already needs Images.jl).
2. We can have a generic utility that can be used with mnist, cifar, imagenet data etc.

ilkerkesen:
OK, but we need to fix hardcoded channel size.

denizyuret:
We can get this from the particular input array can't we?  Also, don't you
need to use colorview to get an image output?

ilkerkesen:
OK, I'll close this PR and I use colorview in plotting procedures. Sometimes I want to save the generations, I am not sure whether colorview breaks saving or not.

denizyuret:
You don't have to use colorview, I just did not know you could display
stuff without converting with colorview...

PR #241, opened 2017-12-25
Add DCGAN Example and Demo

ilkerkesen:
This PR submits a DCGAN (deep convolutional generative adversarial network) example with demo to examples.


denizyuret:
Thanks, this demo seems to work, how did you fix it?

ilkerkesen:
I applied batch normalization after every layer except final layer and it
solved my issues.

PR #242, opened 2017-12-26
Make image grid procedure generic

ilkerkesen:
Make imagegrid works for both RGB and grayscale images. Examples that uses this procedure has been updated.


Evizero:
Just out of curiosity, what does the imagegrid do? is it similar to https://github.com/JuliaArrays/MosaicViews.jl ?

ilkerkesen:
@Evizero Yes, its functionality is very similar to that particular package. I did implement this function months ago, just adapted for Knet recently. That package seems very handy, but it is an additional package and I think `make_image_grid` procedure satisfies our needs currently.

PR #246, opened 2017-12-28
Logistic Loss

ekinakyurek:
I have added two versions of logistic loss to loss.jl. I may also define it as primitive in the future by explicitly writing its derivative. 

Best
Ekin


CarloLucibello:
Maybe you want to add an `average` keyword argument, as in `nll`.

Also, in writing some code for a [variational autoencoder](https://github.com/denizyuret/Knet.jl/blob/ffea2e180f8b605ec19242037434ce6a84693437/examples/variational-autoencoder/vae_mnist.jl#L30), I add to insert a small additive factor inside the logarithms, otherwise I would get  nans and infs after a few epochs

CarloLucibello:
the following method could be defined as well
```
binary_cross_entropy(yhat, y) = logisticloss(yhat,y; mode=2)
```

ekinakyurek:
I added your suggestions. May you test the function in your auto-encoder example? If you got Infs or NaNs, I will also put small additions into the logarithms.

ekinakyurek:
#361

PR #263, opened 2018-01-26
add elu and selu activations

CarloLucibello:


CarloLucibello:
this should be ready for merge after review

cangumeli:
Shouldn't elu and selu be primitives if we add them to Knet?

denizyuret:
You mean Autograd primitives? It depends. If they are defined generically
using other primitives, no need. If they are type specific and/or there are
performance concerns then yes.

cangumeli:
If `elu` and `selu` will be added after each weight layer similar to other activations, defining them as composite operations may bring performance and memory overhead in training. I think they should be implemented similar to how `sigm`, `relu` etc. are implemented.

src/unary.jl
```diff
@@ -148,16 +148,41 @@ import Base: tanh
...
+Self-Normalizing Neural Networks
+https://arxiv.org/abs/1706.02515
+"""
+function selu(x)
+    alpha = Float32(1.6732632)
+    scale = Float32(1.0507009)
+    p = relu(x)
+    m = -relu(-x)
+    return scale*(p + alpha*(exp(m) - 1))
+end
```
quaertym: How about using `elu` here?
```julia
function selu(x)
    alpha = Float32(1.6732632)
    scale = Float32(1.0507009)
    return scale * elu(x, alpha)
end 
```

rened:
Elu can be included in `unary.jl` by adding
`("elu", "elu", "(xi>0?xi:exp(xi)-1)"),`
and
`(:elu, :eluback, :(ifelse(xi>0,xi,exp(xi)-1)), :(ifelse(yi>0,dyi,yi+1))),`
below the respective `relu` lines.

src/broadcast.jl
```diff
@@ -32,6 +32,7 @@ broadcast_ops = [
     # "fdim",
     ("invxback","invxback","(-xi*yi*yi)"),
     ("reluback","reluback","(yi>0?xi:0)"),
+    ("eluback", "eluback", "ifelse(yi>0,dyi,yi+1)"),
```
ozanarkancan: This part turns to a cuda code snippet. So, I think it must be replaced with the following:
```
("eluback", "eluback", "(yi>0?xi:yi+1)")
```

denizyuret:
Those snippets are not Julia code they are Cuda.

denizyuret:
Added selu as a cuda kernel for efficiency.

denizyuret:
@CarloLucibello can you explain how this gives the intended result in alpha_dropout:
` x = q*dropout(x .- alpha, p) .+ alpha #set dropped input to alpha`

PR #274, opened 2018-02-17
CPU and macOS fix

cangumeli:
1. CUDA API could not find clang in my computer (running macOS Sierra). I fixed this by setting `CXX=/usr/bin/clang` if no CXX found, device is apple and clang exists.

2. `find_cuda_library` returns `nothing` if library not found, as opposed to previously used `find_library` which was throwing an exception. Return value being `nothing` causes `@cuda` to throw `ERROR: LoadError: TypeError: ccall: expected Symbol, got Void`. I fixed this by throwing an exception when `path==nothing`.

Knet is now being successfully installed to my laptop which lacks CUDA libraries.


denizyuret:
Thanks.  I took your fix for #2 and opened an issue to CUDAapi for #1.

PR #280, opened 2018-03-03
remove deprecated folders

CarloLucibello:
time to let go... they will always be a few git commands away


denizyuret:
Can you guys let me keep these?  (1) They do not take up extra space.  (2) I can hide them by renaming them .deprecated if you want.  Whenever we move from one major Julia/Knet version to the next, some code breaks, I move them to deprecated, then move them back once things are fixed, i.e. these are not garbage for ever, they are just waiting to be re-polished...

denizyuret:
ok, I renamed them .deprecated so they won't appear when you ls.

Issue #288, opened 2018-03-20
Batchnorm backward on cpu doesn't seem to take care about scales


AndreeaTT:
In batchnorm4_back function derivate(dy) is multiplied with g(scales), but dyivar is used to compute dx derivate. 

 if training
        mu, ivar = _get_cache_data(cache, x, eps)
        x_mu = x .- mu
        # equations from the original paper
        dyivar = dy .* ivar
        if g !== nothing
            dg = sum(x_mu .* dyivar, dims)
            db = sum(dy, dims)
            dy = g .* dy
        else
            dg, db = nothing, nothing
        end
        m = prod(size(x, dims...))
        dsigma2, dmu = similar(ivar), similar(mu)
        dsigma2 .= -T(0.5) .* sum(dyivar .* x_mu .* ivar.^2, dims)
        dmu .= -sum(dyivar, dims) .- 2dsigma2 .* sum(x_mu, dims) ./ m
dx .= dyivar .+ dsigma2 .* 2x_mu ./ m .+ dmu ./ m

CarloLucibello:
@AndreeaTT good catch! I fixed the bug in #289

cangumeli:
Backward equations are fixed thanks to @CarloLucibello's PR. @AndreeaTT thank you for your catch.

It will be really useful if we have a test in [here](https://github.com/denizyuret/Knet.jl/blob/master/test/batchnorm.jl) that consistently fails due to this bug. I thought this would happen if I initialize gamma and beta randomly in tests, but failures only occurred occasionally just in low precision numbers (Float32). We may keep this issue open until such tests are introduced.

PR #289, opened 2018-03-20
fix bnorm cpu back

CarloLucibello:
fix bug #288 in batch normalization cpu back @cangumeli


CarloLucibello:
@denizyuret this is somewhat urgent, since our current implementation of batch norm is broken on gpu

denizyuret:
@cangumeli can you add a test that fails in current master and passes in
new version.

cangumeli:
I think the reason why tests are passing is that `bnparams` function initialially set scales parameters to 1, so input derivative is always computed correctly. I couldn't yet create a consistently failing test by randomly initializing `bnparams`, I will investigate the `gradcheck`s further tomorrow. 

However, as seen in the backward equations at https://arxiv.org/pdf/1502.03167.pdf, `dy` is used instead of `dxhat` in the current version of the code, which is clearly a bug.

Issue #292, opened 2018-03-31
missing digamma kernel for KnetArray


CarloLucibello:
`digamma` is needed if we want to take derivatives of `gamma` and `lgamma`. See also  #290

CarloLucibello:
>Is it possible to adapt this code? https://github.com/rachtsingh/ibp_vae/tree/master/src/lgamma/src

Definitely. I'm no expert here, but I think it could be enough to add a `digamma_impl` here https://github.com/denizyuret/Knet.jl/blob/master/deps/cuda1.jl

and then add to the `unary_ops` list in `unary.jl` the entry
```
("digamma_impl", "digamma", "digamma_impl(xi)"
```

xukai92:
Thanks. How is it possible for me to take derivative of gamma family when not doing broadcasting? Maybe it's on CPU?

CarloLucibello:
On cpu you can compute derivatives even when broadcasting
``` julia
julia> loss(x) = sum(gamma.(x))
loss (generic function with 1 method)

julia> grad(loss)(rand(10))
10-element Array{Float64,1}:
   -9.66163 
   -5.84301 
 -286.216   
  -21.1938  
   -0.907866
   -5.5372  
   -1.23051 
  -16.4395  
   -3.84932 
  -36.5713  
```

Issue #290, opened 2018-03-31
gamma function causes deadlock for broadcasting KnetArray


xukai92:
The code below causes deadlock.

```julia
using Knet
t = KnetArray(ones(10, 10))
gamma.(t)
```

How to resolve this? Basically I need `gamma` family functions to support broadcasting.

CarloLucibello:
I pushed a PR request fixing the deadlock issue, but I fear you will soon hit a bigger problem: derivatives of `gamma` are expressed in term of the `digamma` function, which is unfortunately not in the cuda api, so a specific kernel should be written if one wants to backpropagate

xukai92:
Thanks for the answer! Why is it possible to backprop throw non-broadcasted digamma in AutoGrad.jl?

PR #294, opened 2018-03-31
gamma, lgamma, digamma work with GPU

xukai92:
Related issues:
https://github.com/denizyuret/Knet.jl/issues/290
https://github.com/denizyuret/Knet.jl/issues/292

Related PR:
https://github.com/denizyuret/Knet.jl/pull/291


deps/cuda1.jl
```diff
@@ -3,6 +3,244 @@
 fp = open("cuda1.cu","w")
 using Knet: unary_ops
 
+
+# Digamma
```
CarloLucibello: better put all these functions at the end of the file

CarloLucibello:
in order to pass tests, we should: 
- add a 'REQUIRE` file in the `test` directory containing the single line `SpecialFunctions`. Even better, I'd like SpecialFunctions to be a proper dependence of the package, if @denyizuret agrees. It is a lean and well mantained package
- define the derivative of trigamma(x) as poligamma(2,x) or just avoid testing testing its grad in `test/unary.jl`

.

deps/cuda1.jl
xukai92: I can do this. Or is there a way to put all the codes in a single file?

xukai92:
How should I "define the derivative of trigamma(x) as poligamma(2,x)" in Knet.jl? I think that's in AutoGrad.jl isn't it?

CarloLucibello:
hmhm, looks like `trigamma` derivative is already defined in autograd, but `poligamma(2,x)` has to be defined for KnetArray

denizyuret:
We can make SpecialFunctions a requirement for AutoGrad.

CarloLucibello:
@xukai92 if you don't need `trigamma`'s derivative for KnetArray, the easier way out (once you fix the problem with importing SpecialFunctions) is to just skip the test in  `test/unary.jl`
```julia
@testset "unary" begin
    broken_grads = [trigamma] 
    for f in unary_fns
        f in broken_grads && continue
        ....
```

xukai92:
@CarloLucibello Just come back to work on this PR. I merged with master and it seems everything passes (even I didn't exclude the test as suggested by you long time ago).

xukai92:
I didn't get errors when I run the complete test on my local. Looking at you gist, it seems that some outputs seem to correct but still reported? E.g.

```julia
(f, t, n) = (SpecialFunctions.digamma, Float64, (2, 1))
unary: Test Failed at /home/ec2-user/.julia/dev/Knet/test/unary.jl:39
  Expression: isapprox(cy, Array(gy))
   Evaluated: isapprox([-1.8009; -1.25649], [-1.8009; -1.25649])
Stacktrace:
 [1] macro expansion at /home/ec2-user/.julia/dev/Knet/test/unary.jl:36 [inlined]
 [2] macro expansion at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.0/Test/src/Test.jl:1083 [inlined]
 [3] top-level scope at /home/ec2-user/.julia/dev/Knet/test/unary.jl:6
```

PS: I'm not very familiar with Knet's test scripts.

xukai92:
Oh yes - I also have these errors. We should I do?

xukai92:
Should I set separate `rtol` for these functions or something?

denizyuret:
We can fix rtol, but what about the gradient errors?  They look more
serious, or do you not get them?

denizyuret:
In particular Float32 gradients for digamma/trigamma do not work for me:
```
julia> f(x) = sum(digamma.(x))
f (generic function with 1 method)

julia> grad(f)(rand(Float32,1))
1-element Array{Float32,1}:
 6.1312017

julia> grad(f)(ka(rand(Float32,1)))
1-element KnetArray{Float32,1}:
 NaN

julia> grad(f)(ka(rand(Float64,1)))
1-element KnetArray{Float64,1}:
 2.3219245818432324
```

xukai92:
Thanks for pointing this out. I found that I didn't implement the float and double version in two separate functions correctly. I know how to fix that in C but I don't know how to link them to the Julia function in `ops.kl`. 

Suppose I have `gamma_impl_32` and `gamma_impl_64` as the float and double version, how should I link them to the `gamma` function in Julia?

xukai92:
Seems that simply using the double version can handle both.

```julia
julia> f(x) = sum(digamma.(x))
f (generic function with 1 method)

julia> rs_32 = rand(Float32, 1)
1-element Array{Float32,1}:
 0.35235214

julia> grad(f)(rs_32)
1-element Array{Float32,1}:
 9.129284

julia> grad(f)(ka(rs_32))
1-element KnetArray{Float32,1}:
 9.129284

julia> rs_64 = rand(Float64, 1)
1-element Array{Float64,1}:
 0.6063909020012128

julia> grad(f)(rs_64)
1-element Array{Float64,1}:
 3.573492967407559

julia> grad(f)(ka(rs_64))
1-element KnetArray{Float64,1}:
 3.573492967407558
```

Is it OK to do so?

xukai92:
@CarloLucibello Can we merge this?

CarloLucibello:
Where is the code for the kernels coming from? We should credit the author and avoid issues with licenses.

Also, it's not clear what is happening here with the 32/64 issue. Is the kernel being executed at 64 bit precision also when using float 32 knet arrays? If so, we should fix this

xukai92:
> Where is the code for the kernels coming from? We should credit the author and avoid issues with licenses.

It's based on https://github.com/rachtsingh/lgamma, which is again ported from https://bitbucket.org/eigen/eigen/overview. How should I cite it?

> Also, it's not clear what is happening here with the 32/64 issue. Is the kernel being executed at 64 bit precision also when using float 32 knet arrays? If so, we should fix this

Let me try to solve this

deps/gamma.jl
```diff
@@ -0,0 +1,242 @@
+# Gamma family
+# Acknowledgement:
```
xukai92: @CarloLucibello I added a comment to acknowledge the original files. Is it OK?

deps/cuda1.jl
```diff
@@ -3,9 +3,13 @@
 fp = open("cuda1.cu","w")
 #using Knet: unary_ops
 
-function cuda1src(f, j=f, ex="$f(xi)"; BLK=256, THR=256)
+include("gamma.jl")
+print(fp,cuda1gammafamily())
+
+function cuda1src(f, j=f, ex="$f(xi)"; seperate_impl=false, BLK=256, THR=256)
```
xukai92: @denizyuret I amended the `cuda1src` function to support functions which have different CUDA kernel implementation for float and double. Before the kernel is assumed to be same. Does it look good?

xukai92: @CarloLucibello This float v.s. double issues is solved by this.

xukai92:
@CarloLucibello I resolved the comments we had before. Please see my notes on changes.

deps/gamma.jl
```diff
@@ -0,0 +1,242 @@
...
+    return digamma_impl_$F(x);
+  }
+
+  // dumb code to calculate factorials
+  $T factorial = $one_str;
+  for (int i = 0; i < n; i++) {
+    factorial *= (i + 1);
+  }
+
+  $T s = n % 2 == 0 ? -$one_str : $one_str;
```
xukai92: Just for people who wonder why we had numerical errors before: this line was coded as `s = powf(-1.0, n + 1)` in the original code for float. This causes numerical errors somehow. I changed it to avoid using unnecessary power function to avoid this.

xukai92:
@CarloLucibello Can you take a look at the my updates and see if it's good now? Thanks!

test/REQUIRE
```diff
@@ -0,0 +1 @@
+SpecialFunctions
```
CarloLucibello: I think we don't need this, since SpecialFunctions is already in the REQUIRE of the package

xukai92: Got it.

CarloLucibello:
thanks for your patience and for the good work!

xukai92:
Many thanks for your suggestions on improving codes in this PR!

PR #306, opened 2018-05-17
optimizers for abstract arrays

### LINKED ISSUE #305: Using Knet with StaticArrays
It would be a nice feature to enable the speedup of StaticArrays for Knet if it's not too hard to bring this in. It could result in a nice speedup for training if matrices are not too big, it certainly does for inference.
At the moment, the creation of the optimizer returns a Void object, which leads to errors later when calling "optimize!".
```julia
using Knet, StaticArrays

w1 = [randn(5, 5), randn(5)]
println(optimizers(w1, Adam)) # This works

w2 = [SMatrix{5, 5}(randn(5, 5)), SVector{5}(randn(5))]
println(optimizers(w2, Adam)) # This does not produce an error but returns Void
```

CarloLucibello:
fix #305 (at least the part regarding the call to `optimizers`)

### DIFF
--- a/src/update.jl
+++ b/src/update.jl
@@ -499,9 +499,9 @@ that parallel model parameters easy when all of them use the same type
 and options.
 
 """
-optimizers{T<:Number}(::KnetArray{T},otype; o...)=otype(;o...)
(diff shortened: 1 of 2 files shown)


denizyuret:
Manually merged to master.

PR #308, opened 2018-05-28
resnetlib debug fix

ekinakyurek:
Moment variances were not squared when loaded. Now, it is working correctly.


ekinakyurek:
This mistake also exists in current master. @denizyuret you may fix it manually.

PR #339, opened 2018-08-23
JLD2 interface for Knet

ekinakyurek:
Knet.save, Knet.load, Knet.@save, Knet.@load functions/macros added
serialization for RNN and KnetArray added
deep serialization
test cases

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -11,6 +11,8 @@ LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+JLD2 = "033835bb-8acc-5ee8-8aae-3f567f8a3819"
+FileIO = "5789e2e9-d7fb-5bc7-8068-2c6fae9b9549"
 
 [compat]
 AutoGrad = "1.0.0"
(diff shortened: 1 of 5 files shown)


Project.toml
```diff
@@ -11,6 +11,8 @@ LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+JLD2 = "033835bb-8acc-5ee8-8aae-3f567f8a3819"
+FileIO = "5789e2e9-d7fb-5bc7-8068-2c6fae9b9549"
```
denizyuret: do we need both JLD2 and FileIO?

denizyuret: We may need to update REQUIRE also.

src/Knet.jl
```diff
@@ -35,6 +35,7 @@ include("distributions.jl"); 	export gaussian, xavier, bilinear
 include("random.jl");           export setseed
 include("hyperopt.jl");         export hyperband, goldensection
 include("data.jl");             export minibatch
+include("jld.jl");              export RnnJLD,KnetJLD
```
denizyuret: Shouldn't we call this jld2.jl ?

src/jld.jl
```diff
@@ -0,0 +1,172 @@
+using JLD2, FileIO
+
+struct RnnJLD; inputSize; hiddenSize; numLayers; dropout; inputMode; direction; mode; algo; dataType; end
+struct KnetJLD; a::Array ; end
+
+function save(fname,args...;kwargs...)
+     FileIO.save(fname,serialize.(args)...;kwargs...)
+end
+function load(fname,args...;kwargs...)
+     serialize(FileIO.load(fname,serialize.(args)...;kwargs...))
```
denizyuret: why are we serializing when loading?

src/jld.jl
```diff
@@ -0,0 +1,172 @@
+using JLD2, FileIO
```
denizyuret: ok, explain this code to me... let's do a hangout.

PR #343, opened 2018-08-29
fixed IdDict case in serialization

ekinakyurek:


ekinakyurek:
This fails because current master is also failing.


```JULIA
ERROR: LoadError: LoadError: UndefVarError: Param not defined
Stacktrace:
 [1] top-level scope at none:0
 [2] include at ./boot.jl:317 [inlined]
 [3] include_relative(::Module, ::String) at ./loading.jl:1038
 [4] include at ./sysimg.jl:29 [inlined]
 [5] include(::String) at /builds/JuliaGPU/Knet.jl/src/Knet.jl:1
 [6] top-level scope at none:0
 [7] include at ./boot.jl:317 [inlined]
 [8] include_relative(::Module, ::String) at ./loading.jl:1038
 [9] include(::Module, ::String) at ./sysimg.jl:29
 [10] top-level scope at none:2
 [11] eval at ./boot.jl:319 [inlined]
 [12] eval(::Expr) at ./client.jl:399
 [13] top-level scope at ./none:3
in expression starting at /builds/JuliaGPU/Knet.jl/src/jld.jl:91
in expression starting at /builds/JuliaGPU/Knet.jl/src/Knet.jl:48
```

src/jld.jl
```diff
@@ -166,7 +163,7 @@ function serialize_internal(x::Union{Dict,IdDict}, stackdict::IdDict)
         return (stackdict[x] = x)
     end
 
-    dest = Dict()
+    dest = x <: Dict ? Dict() : IdDict()
```
denizyuret: This may not recover the original type, as there are more cases. Maybe you should store the original type as well?

```
julia> subtypes(AbstractDict)
8-element Array{Any,1}:
 Base.EnvDict
 Base.ImmutableDict
 Base.Iterators.Pairs
 Dict
 IdDict
 OrderedCollections.OrderedDict
 Test.GenericDict
 WeakKeyDict
```

ekinakyurek: I intentionally did this. For example if the input is Dict{String,KnetArray} , we cannot save it as Dict{String,KnetJLD}

denizyuret: I understand.  However when we save/load some data we should recover all the original types.  Thus my suggestion of "store the original types as well", maybe as part of a struct?

denizyuret:
@ekinakyurek Please test with subtypes of AbstractDict to make sure we load the same types that we save.

PR #352, opened 2018-09-11
julia1 compat for rl examples

ozanarkancan:


denizyuret:
I am making a couple of small changes: (1) PlotlyJS is still not stable, removing that (default uses GR now) (2) Added some automatic installation stuff at the beginning of dp. (3) We can save the notebooks with plot outputs without blowing up the size now (see https://discourse.julialang.org/t/plots-makes-notebooks-too-large-in-ijulia/14550/13), so I am saving them with outputs.

PR #358, opened 2018-09-23
1.0 Compatible Version of #313. Fixes `logp` and adds  `softmax` functionality

ekinakyurek:
I realized that current `logp` does not use cudnn function because of wrong condition. There was a mistake for dims==2 too. I replicated #313 for Julia 1.0 and all problems are solved.

### DIFF
--- a/src/loss.jl
+++ b/src/loss.jl
@@ -11,23 +11,7 @@ given dimensions.  In particular, if `x` is a matrix, `dims=1`
 normalizes columns of `x` and `dims=2` normalizes rows of `x`.
 
 """
-function logp(x;dims=:)
-    if isa(value(x), KnetArray)
-        if dims==1
-            cudnnSoftmaxForward(x,algo=2)
-        elseif dims==2 && ndims(x)==2
-            cudnnSoftmaxForward(x',algo=2)'
-        elseif dims==:;
-            n = length(x)
-            (n > 20000 ? _logp(x) : # see Knet/prof/softmax.jl for timing info
-            reshape(cudnnSoftmaxForward(reshape(x,(1,1,n,1)),algo=2),size(x)))
-        else
-            _logp(x,dims=dims)
-        end
-    else
-        _logp(x,dims=dims) # fall back on old implementation if not KnetArray
-    end
-end
+logp(x;dims=:,algo=2) = generic_softmax(x,algo,_logp;dims=dims)
 
 # Math for the cross-entropy loss: x is unnormalized input, p is
 # target probabilities, q is estimated probabilities. Read left column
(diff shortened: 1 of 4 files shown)


src/loss.jl
```diff
@@ -101,6 +85,56 @@ mutable structdef enum
 
 =#          
 
+
+"""
+    softmax(x;[dims])
```
CarloLucibello: softmax(x; dims =:, algo=1)

test/loss.jl
```diff
@@ -28,5 +28,7 @@ include("header.jl")
         @test isapprox(nll(k, indices, dims=1), nll(a, indices, dims=1))
         @test isapprox(nll(k, indices, dims=2), nll(a, indices, dims=2))
     end
+    logistic(a[:],a[:])
+    bce(a[:],a[:])
```
CarloLucibello: this is not a test

src/loss.jl
```diff
@@ -171,6 +205,31 @@ function nll(y,a::AbstractArray{<:Integer}; dims=1, average=true)
...
+    ε = eltype(x̂)(1e-12)
+    l = log.((1-ε) .+ exp.(-x .* x̂))
+    average ? mean(l) : sum(l)
+end
+
+"""
+
+    bce(scores,answers;average=true)
+
+Computes binary cross entropy given scores(predicted values) and answer labels.
```
CarloLucibello: space after scores

test/loss.jl
CarloLucibello: also, logsoftmax should be tested

src/loss.jl
```diff
@@ -101,6 +85,56 @@ mutable structdef enum
...
+"""
+const logsoftmax = logp
+
+function generic_softmax(x,algo::Int,fallback;dims=:)
+    if isa(value(x), KnetArray)
+        if dims==1
+            sz = size(x)
+            x = cudnnSoftmaxForward(reshape(x, (1,1,sz[1],:)), algo=algo)
+            reshape(x, sz)
+        elseif dims==2 && ndims==2
```
CarloLucibello: ndims -> ndims(x)

An error like this should be caught when tested on a gpu. If not, please add a test

CarloLucibello:
There are some parts of #313  that I don't see here, and some tests also. I think
it is better to simply rebase #313, then merging this as it is. 
We can shrink this pr to just the addition of the bce and logistic losses.
Also, when building upon a stale PR, it would be nice to preserve the authorship of the first commits. It is
not hard to do, just a matter of branching for the stale branch.

src/loss.jl
ekinakyurek: done

src/loss.jl
ekinakyurek: I tried be consistent with master. We may go this interface.

src/loss.jl
ekinakyurek: tests added, typo is corrected.

test/loss.jl
ekinakyurek: logsoftmax is jus an alias. This will increase test time. Do we really need it ?

src/loss.jl
ekinakyurek: Sorry, you can use this functionality, but I didn't expose it in the documentation.

Issue #295, opened 2018-04-04
Feature request: cuBLAS Strided Batched Matrix Multiply


ngphuoc:
This feature would enable networks which use matrices as inputs, states, and outputs. It allows running on mini batches of matrices.

ref: https://devblogs.nvidia.com/cublas-strided-batched-matrix-multiply/

cangumeli:
What kind of Julia interface do you recommend for this operation?

ngphuoc:
I think it is best that the same code can run on both CPU and GPU. I think the syntax in `@tensor` in TensorOperations.jl would be an approach. The second is mapslices (https://github.com/JuliaComputing/ArrayFire.jl/issues/187). And the third option, which may has simplest syntax and need extension overload on CPU, is to extend existing blas operators.

cangumeli:
What about a `bmm` function that simply takes two 3d arrays and perform batch matrix multiplication?

Your third option seems very optimal in terms of syntax, but I'm not sure about changing default behaviour of julia matrix multiplication.

ngphuoc:
Yes, that would be reasonable. Users can override `*` for their personal need. An additional `bmm!(out, batch1, batch2)` would be helpful.

ngphuoc:
And a `btranspose` if it is faster than `permutedims`.

Roger-luo:
Just to mention, I have added a `gemmBatchedStrided` wrapper in CuArrays.jl: https://github.com/JuliaGPU/CuArrays.jl/pull/190

denizyuret:
PR #363 was merged with master. @ngphuoc does this resolve this issue?

ngphuoc:
I works great! Thanks

PR #363, opened 2018-09-26
Batched Matrix Multiplication Operation

ozanarkancan:
Hi,

This is a premature pr for the #295. I simply made an interface for the [gemmStridedBatched](https://devblogs.nvidia.com/cublas-strided-batched-matrix-multiply/) call. All comments are welcome.

```julia
#size of x1: (m,k,bs)
#size of x2: (k,n,bs)
#size of output: (m,n,bs)
output = bmm(x1, x2)
```


src/bmm.jl
```diff
@@ -0,0 +1,63 @@
...
+    end
+    return C
+end
+
+function bmm!(A, B, C)
+    m,n,bs = size(A)
+    n,k,bs = size(B)
+    
+    for i=1:bs
+        C[:,:,i] = reshape(A[:,:,i], m, n) * reshape(B[:, :, i], n, k)
```
ekinakyurek: may be it is good to use `view(A,:,:,i)` here. I think we should have tests for `bmm` with profiling to compare different methods.

ngphuoc:
Thanks a lot for the PR. I got 100 times speedup in the following benchmark!

```julia
using Knet, BenchmarkTools

m,n,p,b = 100,200,300,50
atype = KnetArray{Float32}
a = rand(m,n,b)
b = rand(n,p,b)
a_ = atype(a)
b_ = atype(b)

@btime for i=1:size(a,3)
  a[:,:,i] * b[:,:,i]
end;  # 11.323 ms (651 allocations: 41.98 MiB)

@btime for i=1:size(a,3)
  a_[:,:,i] * b_[:,:,i]
end;  # 338.570 μs (1999 allocations: 64.06 KiB)

@btime c = bmm(a, b);  # 12.012 ms (302 allocations: 53.42 MiB)

@btime c_ = bmm(a_, b_);  # 3.822 μs (7 allocations: 336 bytes)
```julia

File: src/bmm.jl [hunk 2]
```diff
@@ -0,0 +1,68 @@
...
+
+function bmm!(A, B, C)
+    if ndims(A) != 3 || ndims(B) != 3
+        throw(DimensionMismatch("$(map(size,(A,B,C)))"))
+    end
+    ma,ka,bsa = size(A)
+    kb,nb,bsb = size(B)
+
+    (bsa == bsb && ka == kb && size(C,1) == ma && size(C,2) == nb && size(C,3) == bsa) || throw(DimensionMismatch("$(map(size,(A,B,C)))"))
+
```
ngphuoc: I need to add this
```suggestion
    m,n,k,bs = ma,nb,ka,bsa
```

denizyuret:
I assume this is ready to go?

ozanarkancan:
Yes, this can be merged.

ilkerkesen:
Resolved the merge conflict and fixed cuda macro call. Tests pass but I don't now anything about the table in `runtests.jl`. It's working and ready to be merged. @denizyuret

PR #365, opened 2018-09-27
Use Plots#master in tutorials

bovine3dom:
The Plots.jl stable release is currently broken on Julia v1.0. This PR makes the tutorials install the latest master branch of Plots. The PR can be reverted once https://github.com/JuliaPlots/Plots.jl/issues/1631 is closed.

Without this PR, the notebooks don't work out of the box on Julia v1.0.

(Thanks for the lovely tutorials, by the way.)

### DIFF
--- a/tutorial/00.Julia_is_fast.ipynb
+++ b/tutorial/00.Julia_is_fast.ipynb
@@ -146,7 +146,7 @@
    "outputs": [],
    "source": [
     "using Pkg\n",
-    "for p in (\"BenchmarkTools\",\"Plots\",\"PyCall\",\"Conda\")\n",
+    "for p in (\"BenchmarkTools\",PackageSpec(name=\"Plots\",rev=\"master\"),\"PyCall\",\"Conda\")\n",
     "    haskey(Pkg.installed(),p) || Pkg.add(p)\n",
     "end"
    ]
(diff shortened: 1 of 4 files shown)


denizyuret:
Hi Oliver,

Thank you. I have been using the gr() backend which seems to work. I could
not get the plotlyjs() backend working, but the default settings seem to be
ok on my linux box and on colab. In which environment are the notebooks
broken in your experiments?

deniz

bovine3dom:
I was getting an error at `using Plots`, as per https://github.com/JuliaPlots/Plots.jl/issues/1633. I didn't get a chance to choose a backend. GR works fine after installing Plots#master - other backends untested.

I'm using:
```
julia --version: julia version 1.0.0
uname -a: Linux 4.17.12-arch1-1-ARCH #1 SMP PREEMPT  x86_64 GNU/Linux
Manifest.toml: [[Knet]] version = "1.1.0"
```

Thanks.

ekinakyurek:
Adding `#master` may cause problems in later. I also do not get any error with GR backend in the current Plots release.

ekinakyurek:
Could you try again and report the error you get ?

bovine3dom:
I was getting exactly the same error as is detailed in the plots issue I linked to above.

I agree linking to master is brave. Perhaps we can pin it to a specific commit?

I'm happy to try again when I am at my PC tomorrow.

bovine3dom:
I've left it long enough that the issue has gone away - Plots v0.21.0 works fine for me now.

Closing this PR as it is therefore not needed. Thanks!

PR #374, opened 2018-10-09
Fix bce logistic for KnetArrays

ekinakyurek:
This PR will result with an error on `convert` function. I think there is a possible bug on `gradcheck` because it boxes second argument too, or, there may be bug in convert,  because it cannot convert `Param{Array{Float64}}` to `Param{KnetArray{Float64]}`.

```JULIA
[26] _start() at ./client.jl:421loss: Error During Test at /home/gridsan/eakyurek/gitother/Knet.jl/test/loss.jl:61
  Test threw exception
  Expression: gradcheck(logistic, k[:, 1], ones(10))
  MethodError: Cannot `convert` an object of type Param{Array{Float64,1}} to an object of type Param{KnetArray{Float64,1}}
  Closest candidates are:
    convert(::Type{T}, !Matched::T) where T at essentials.jl:154
  Stacktrace:
   [1] #differentiate#1(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::Function, ::Function) at /home/gridsan/eakyurek/.julia/packages/AutoGrad/eAmjh/src/core.jl:57
   [2] differentiate at /home/gridsan/eakyurek/.julia/packages/AutoGrad/eAmjh/src/core.jl:45 [inlined]
   [3] #gradcheck#3(::Tuple{}, ::Function, ::Int64, ::Int64, ::Float64, ::Float64, ::Float64, ::Function, ::typeof(logistic), ::KnetArray{Float64,1}, ::Vararg{Any,N} where N) at /home/gridsan/eakyurek/.julia/packages/AutoGrad/eAmjh/test/gradcheck.jl:39
   [4] gradcheck(::Function, ::KnetArray{Float64,1}, ::Vararg{Any,N} where N) at /home/gridsan/eakyurek/.julia/packages/AutoGrad/eAmjh/test/gradcheck.jl:36
   [5] macro expansion at /home/gridsan/eakyurek/gitother/Knet.jl/test/loss.jl:61 [inlined]
   [6] macro expansion at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.0/Test/src/Test.jl:1083 [inlined]
   [7] top-level scope at /home/gridsan/eakyurek/gitother/Knet.jl/test/loss.jl:4
```

### DIFF
--- a/src/loss.jl
+++ b/src/loss.jl
@@ -226,9 +226,8 @@ Computes logistic loss given scores(predicted values) and answer labels.
 answer values should be {-1,1}, then it returns `mean|sum(log(1 + exp(-answers*scores)))`. See also `bce`.
 """
 function logistic(x̂,x;average=true)
-    ε = eltype(x̂)(1e-12)
-    x = convert(typeof(x̂),x)
-    l = log.((1-ε) .+ exp.(-x .* x̂))
+    x = oftype(value(x̂),x)
+    l = log.(1 .+ exp.(-x .* x̂))
     average ? mean(l) : sum(l)
 end
 
@@ -243,8 +242,8 @@ where `p` is equal to `1/(1 + exp.(scores))`. See also `logistic`.
 function bce(x̂,x;average=true) 
     ε = eltype(x̂)(1e-12)
     p = 1 ./ (1 .+ exp.(-x̂))
-    x =	convert(typeof(x̂),x)
-    l = x .* log.(p .+ ε) .+ (1 .- x).*log.((1-ε) .- p)
+    x =	oftype(value(x̂),x)
+    l = x .* log.(min.(p,ε)) .+ (1 .- x).*log.(1 .- max.(p,1-ε))
     average ? -mean(l) : -sum(l)
 end
 
--- a/src/loss.jl
+++ b/src/loss.jl
@@ -227,6 +227,7 @@ answer values should be {-1,1}, then it returns `mean|sum(log(1 + exp(-answers*s
 """
 function logistic(x̂,x;average=true)
     ε = eltype(x̂)(1e-12)
+    x = convert(typeof(x̂),x)
     l = log.((1-ε) .+ exp.(-x .* x̂))
     average ? mean(l) : sum(l)
 end
@@ -242,6 +243,7 @@ where `p` is equal to `1/(1 + exp.(scores))`. See also `logistic`.
 function bce(x̂,x;average=true) 
     ε = eltype(x̂)(1e-12)
     p = 1 ./ (1 .+ exp.(-x̂))
+    x =	convert(typeof(x̂),x)
     l = x .* log.(p .+ ε) .+ (1 .- x).*log.((1-ε) .- p)
     average ? -mean(l) : -sum(l)
 end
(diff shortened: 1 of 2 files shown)


src/loss.jl
```diff
@@ -243,8 +242,8 @@ where `p` is equal to `1/(1 + exp.(scores))`. See also `logistic`.
 function bce(x̂,x;average=true) 
     ε = eltype(x̂)(1e-12)
     p = 1 ./ (1 .+ exp.(-x̂))
-    x =	convert(typeof(x̂),x)
-    l = x .* log.(p .+ ε) .+ (1 .- x).*log.((1-ε) .- p)
+    x =	oftype(value(x̂),x)
+    l = x .* log.(min.(p,ε)) .+ (1 .- x).*log.(1 .- max.(p,1-ε))
```
ekinakyurek: ?```l = x .* log.(max.(p,ε)) .+ (1 .- x).*log.(1 .- min.(p,1-ε))```

denizyuret: That's right, sorry about that.  I have a possibly stupid question: isn't logistic and bce doing pretty much the same thing (one with {-1,1} the other with {0,1}?).  If that is correct why do we need epsilons in one and not the other?  Can't we just convert one gold to the other?

denizyuret:
My latest thoughts: 
* our convention is to leave the last layer of a network linear, and do normalization, softmax etc in the loss function.
* from that perspective bce and logistic compute exactly the same thing so we don't need them.
* instead of introducing a new function we could have nll do this calculation if the model output is 1-D.
* it would be a lot faster to add a simple cuda kernel for the 1-D nll calculation.

ekinakyurek:
My thoughts:

* Do not mix users mind with nll and bce/logistics together. Here, I assume, users will have a clear context for bce and they will have separate context for nll.
* Rename logistic->bce, and delete current bce.
* Write gpu kernel for bce.

Additionally,

* add alias for nll as crossentropy
* add alias for bce as binarycrossentropy

denizyuret:
I was trying to unconfuse the users, not to further confuse them: binary
cross entropy IS negative log likelihood when there is a single score used
to compute probabilities for two classes. I think the general tendency to
give the same thing 5 different names for historical reasons is the
confusing thing - it prevents people from understanding the common
concept.  However I agree that in this case we can keep bce separate.

ekinakyurek:
I unified bce and logistic, we may merge this now. After writing a gpu kernel, we can update again.

ekinakyurek:
can be merged for now?

test/loss.jl
```diff
@@ -54,13 +54,15 @@ include("header.jl")
     indices = rand(1:10,10)
     @test gradcheck(nll, a, indices, kw=(:dims=>1,), args=1)
     @test gradcheck(nll, a, indices, kw=(:dims=>2,), args=1)
+    @test gradcheck(logistic,a[:,1],ones(10))
+    @test gradcheck(bce,a[:,1],ones(10))
     if gpu() >= 0
         k = KnetArray(a)
+	@test gradcheck(logistic,k[:,1],ones(10))
+    	@test gradcheck(bce,k[:,1],ones(10))
```
CarloLucibello: indentation

denizyuret:
bce/logistic are in src/ops20/loss.jl in 1.4.1.

PR #396, opened 2018-11-21
using CuArrays

MikeInnes:
The idea of this is to gradually start sharing more code with CuArrays. For now it essentially just replaces Knet's allocation code (Kptr) with a `CuArray`. With that + appropriate conversion to pointers, Knet's existing GPU kernels work, and around 90% of the tests pass already.

My hope is that we can get the tests passing here, check that any performance issues are resolved, and anything else that's needed to make this mergeable. Then we can follow up with PRs that e.g. replace the CUBLAS and CUDNN wrappers with straightforward CuArrays calls, enable broadcast fusion, etc.

TODO:

* [x] Figure out getindex issue
* [ ] Enable RNN tests
* [ ] Update JLD integration


MikeInnes:
With some more updates almost all the tests pass. I'm getting a strange issue with gradchecks on `getindex` (despite the fact that KnetArray itself behaves the same as before, and I haven't touched any AD code), so that needs looking into. I think this is ready for some review and testing though, to make sure the approach is sound.

MikeInnes:
Only temporarily – my hope is that you'll experiment with it, and once you're happy there's no perf loss we'll merge it.

It might need some collaboration to push it through if there are issues on the Knet side, but I'm certainly happy to help fix anything on the CuArrays side.

maleadt:
How is the allocation performance? CuArray's allocator hasn't been tuned yet, except for the large buffer opt-out.

MikeInnes:
That would be my ideal eventually, but for now I'm trying to preserve Knet's current semantics as much as possible, including things like the current subtyping, logical indexing and broadcasting (which can't co-exist with CuArrays' version).

It's also a bit non-trivial because KnetArray goes a bit beyond the array interface. E.g. the JLD code looks inside the `Kptr` and replaces the GPU `Ptr` with a full `Array` of values for serialisation. So the full refactor is probably quite a large project, but the benefit is that Knet will then be able to work with `CLArray`, `XRTArray` on TPUs etc.

denizyuret:
Found it: your modified setindex! returns the value rather than the array at src/karray.jl:413. If you fix that getindex tests should pass.

denizyuret:
> That would be my ideal eventually, but for now I'm trying to preserve Knet's current semantics as much as possible, including things like the current subtyping, logical indexing and broadcasting (which can't co-exist with CuArrays' version).
> 
> It's also a bit non-trivial because KnetArray goes a bit beyond the array interface. E.g. the JLD code looks inside the `Kptr` and replaces the GPU `Ptr` with a full `Array` of values for serialisation. So the full refactor is probably quite a large project, but the benefit is that Knet will then be able to work with `CLArray`, `XRTArray` on TPUs etc.

The JLD code is an isolated hack: JLD was broken, and JLD2 did not allow type specific load/save methods, so I did this ptr hack to avoid dealing with changing types. So separating the code should not be that difficult. I plan to separate KnetArray into its own package and making Knet array-type agnostic at some point. However this will put all the kernels and CuDNN calls into the KnetArray package and not necessarily help CuArrays use Knet kernels, unless you guys suggest some other way of organizing things...

MikeInnes:
I'm now getting several errors of the form:

```julia
linalg: Error During Test at /home/mikeinnes/projects/Knet/test/linalg.jl:19
  Test threw exception
  Expression: gradcheck(mmul, (ka, kb))
  ccall: could not find function cudaGetErrorString in library /usr/local/cuda-9.1/lib64/libcublas.so
```

I guess there's a real error here but it's getting obscured by CUDA error handling also not working.

denizyuret:
All tests pass except rnn and jld.  In rnn we seem to have a problem with reshaping shared KnetArray: i.e. `k = ka(rand(3,5)); reshape(k[1:4], (2,2))` which happens during RNN initialization: in the new implementation the k[1:4] does not give a shared array.

denizyuret:
Issue that prevents rnn from working: transpose does not work on views: (k[4:12] below returns a view)
```
julia> Base.getindex(x::KnetArray, r::UnitRange) = KnetArray(view(x.ptr, r))
julia> a = rand(3,5)
a = rand(3,5)
3×5 Array{Float64,2}:
 0.598898  0.337149   0.225076   0.57478   0.664271
 0.366906  0.534062   0.0599002  0.691382  0.115686
 0.290002  0.0321779  0.0910531  0.732437  0.21513

julia> k = ka(a)
k = ka(a)
3×5 KnetArray{Float64,2}:
 0.598898  0.337149   0.225076   0.57478   0.664271
 0.366906  0.534062   0.0599002  0.691382  0.115686
 0.290002  0.0321779  0.0910531  0.732437  0.21513

julia> s = reshape(k[4:12],(3,3))
s = reshape(k[4:12],(3,3))
3×3 KnetArray{Float64,2}:
 0.337149   0.225076   0.57478
 0.534062   0.0599002  0.691382
 0.0321779  0.0910531  0.732437

julia> s'
s'
3×3 KnetArray{Float64,2}:
 0.598898  0.366906   0.290002
 0.337149  0.534062   0.0321779
 0.225076  0.0599002  0.0910531
```

denizyuret:
Now with a few more fixes I can get rnn's to pass but test/karray.jl is failing.

MikeInnes:
closing in favour of #417

PR #419, opened 2018-12-18
broadcasting user defined functions

denizyuret:


CarloLucibello:
Hi, thank you for this!

Autograd seems to behave badly with broadcasting also with standard arrays (cf. https://github.com/denizyuret/AutoGrad.jl/issues/103)

Do we get back the gradient of user defined functions only when using knetarrays with this pr ?

PR #421, opened 2018-12-21
Update reference.md

ahmetumutdurmus:
I believe `rnninit` and `rnnforw` have been switched to `RNN` and `Sgd` has been renamed as `SGD`. The docs at: http://denizyuret.github.io/Knet.jl/latest/reference.html#Recurrent-neural-networks-1 and http://denizyuret.github.io/Knet.jl/latest/reference.html#Optimization-methods-1 are not displaying these.


denizyuret:
Thank you.

Issue #321, opened 2018-07-20
Knet's vcat causing ambiguity with Base 


CarloLucibello:
Knet introduce an imbiguity with base method `vact` when working with arrays with dimension greater than two and trying to `vcat` more than 2 arrays. 
```julia
using Knet 
julia> x = rand(2,2,2,2)
julia> vcat(x,x,x)
ERROR: MethodError: vcat(::Array{Float64,4}, ::Array{Float64,4}, ::Array{Float64,4}) is ambiguous. Candidates:
  vcat(a::Union{AbstractArray, Number}, as::Union{AbstractArray, Number}...) in Knet at /home/lucibello/.julia/v0.6/Knet/src/karray.jl:300
  vcat(a::Union{AbstractArray, Knet.KnetArray, Number}, as::Union{AbstractArray, Knet.KnetArray, Number}...) in Knet at /home/lucibello/.julia/v0.6/Knet/src/karray.jl:301
  vcat(A::AbstractArray...) in Base at abstractarray.jl:1311
Possible fix, define
  vcat(::AbstractArray, ::Vararg{AbstractArray,N} where N)
```
`vcat` with just 2 array or with arrays with dimension 2 is fine.

PR #422, opened 2019-01-05
Fixing ambiguity issue #321

denizyuret:


CarloLucibello:
I wonder if #321 could be fixed avoiding type piracy. After this PR we have this:
```julia
# Avoid using Base for unimplemented cat methods:
using AutoGrad: NA # Union{Number,AbstractArray}
const NAK = Union{Number,AbstractArray,KnetArray}
cat(a::NA, as::NA...; dims)=Base._cat(dims, a, as...)
cat(a::NAK, as::NAK...; dims)=throw(MethodError(cat, (a, as...)))
hcat(a::AbstractArray, as::AbstractArray...)=cat(a,as...; dims=2) # ambiguity fix #321
hcat(a::NA, as::NA...)=cat(a,as...; dims=2)
hcat(a::NAK, as::NAK...)=throw(MethodError(hcat, (a, as...)))
vcat(a::AbstractArray, as::AbstractArray...)=cat(a,as...; dims=1) # ambiguity fix #321
vcat(a::NA, as::NA...)=cat(a,as...; dims=1)
vcat(a::NAK, as::NAK...)=throw(MethodError(vcat, (a, as...)))
```
Looks like deleting all of these methods would fix #321. What would we lose?

Issue #344, opened 2018-08-31
JLD2 vs Knet load behave differently


denizyuret:
When loading [this file](http://people.csail.mit.edu/deniz/data/mnist.jld2) we seem to have a problem:
```
using JLD2, FileIO
a1 = FileIO.load("mnist.jld2","dtst")
a2 = Knet.load("mnist.jld2","dtst")
first(a1) => ok
first(a2) => error
```
As far as I can tell, a1 and a2 are identical except for the xtype field.  
```
for n in fieldnames(Knet.Data)
    println(n,"\t",getfield(a1,n)==getfield(a2,n))
end
x	true
y	true
batchsize	true
length	true
partial	true
indices	true
shuffle	true
xsize	true
ysize	true
xtype	false
ytype	true
```
Both a1.xtype and a2.xtype print the same but they seem to be not equal! If I replace a2.xtype = a1.xtype it starts working.  You may want to check how types are saved...

ekinakyurek:
I actualy do nothing for types, it is very strange. I could not see why this is happening. Do you have any idea?

ekinakyurek:
this issue is still exist, very interesting

ekinakyurek:
I think @JeffBezanson may have a comment on that.

JeffBezanson:
Is the TypeVar N === to the one inside the Array type in the body field?

ekinakyurek:
```JULIA
julia> a1.xtype.var == a2.xtype.var
false

julia> a1.xtype.var
N

julia> a2.xtype.var
N
```

ekinakyurek:
We were replicating deepcopy in our saving code. So, I found this JuliaLang/julia#30911

PR #433, opened 2019-02-02
fix serialize bug in #344

ekinakyurek:
After Julia fixed deepcopy bug, I fix our serialization bug too.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -15,3 +15,6 @@ include:
 
 test:dev:
   allow_failure: true
+
+coverage:
+  allow_failure: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,10 @@ julia:
   - 1.0
   - nightly
 
+matrix:
+  allow_failures:
+    - julia: nightly
+
 notifications:
   email: false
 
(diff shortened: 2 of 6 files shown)


denizyuret:
#425 ve #433 icin simdi fail eden duzeltmeden sonra gecen test lazim...

ekinakyurek:
I merged with #425 and added the tests that fails before

ekinakyurek:
@denizyuret could you try below by activating this project https://github.com/ekinakyurek/Morse.jl/tree/predict on a cpu only machine, which requires internet connection to download the saved model:

```JULIA
julia> using Knet, KnetLayers , Morse
julia> model, vocabulary, parser = trained(MorseModel, TRDataSet, vers="2018");
julia> predictions = model("annem sana yardım edemez .", v=vocabulary, p=parser)
```

I'm getting assertion error on size of hidden states, whereas I don't get any error on a GPU machine.

denizyuret:
> ```julia
> ```julia
> julia> using Knet, KnetLayers , Morse
> julia> model, vocabulary, parser = trained(MorseModel, TRDataSet, vers="2018");
> julia> predictions = model("annem sana yardım edemez .", v=vocabulary, p=parser)
> ```
> ```

You had a 2-D hidden state instead of 3-D. I modified the code so the assertion ignores trailing 1 dimensions.

PR #436, opened 2019-02-15
Print ETA

ekinakyurek:
I believe here we print total time, instead we should print ETA.


denizyuret:
What do you mean “SHOULD”?

I didn’t like curr<eta, so I replaced it with curr/total.

ekinakyurek:
Opps, I assume you meant to print ETA but accidently print Total. That’s why I said should.

Although it is not important, however, original tqdm is also prints curr<eta which I more like.
https://pypi.org/project/tqdm/ <https://pypi.org/project/tqdm/>

my logic is: 
-You wonder how much time remained to complete this loop
-You can see how much time taken in the end

ekinakyurek:
- also, addition is  generally easier than subtraction :)

denizyuret:
:)

I will think about it.

PR #445, opened 2019-03-12
Fix typo in LeNet Chain call

Alexander-Barth:
I get the following error:

```
ERROR: LoadError: MethodError: no method matching Chain(::Conv, ::Conv, ::Dense, ::Dense)
Closest candidates are:
  Chain(::Any) at /mnt/data1/abarth/work/kaggle/cell_images/classification_knet.jl:14
```

when running the MNIST example. Creating a tuple (or an array) of layers work.


denizyuret:
Thank you.

PR #466, opened 2019-06-17
Add Masked NLL interface

ilkerkesen:
This PR adds Masked NLL functionality with two different interfaces. Currently, tests fail because of some multiple dispatch pattern matching issue.

### DIFF
--- a/src/loss.jl
+++ b/src/loss.jl
@@ -84,7 +84,7 @@ mutable structdef enum
     CUDNN_SOFTMAX_MODE_CHANNEL = 1     /* compute the softmax over all C for each H, W, N */
 } cudnnSoftmaxMode_t;
 
-=#          
+=#
 
 
 """
@@ -92,10 +92,10 @@ mutable structdef enum
     softmax(x; dims=1, algo=1)
 
 The softmax function typically used in classification.
-Gives the same results as to `exp.(logp(x, dims))`. 
+Gives the same results as to `exp.(logp(x, dims))`.
 
-If `algo=1` computation is more accurate, if `algo=0` it is 
-faster. 
+If `algo=1` computation is more accurate, if `algo=0` it is
+faster.
 
 See also `logsoftmax`.
 
@@ -108,7 +108,7 @@ function _softmax(x; dims=:, algo=1)
     @assert algo ∈ [0, 1]
     if algo == 1
         x = x .- maximum(x, dims=dims)
-    end    
+    end
     x = exp.(x)
     return x ./ sum(x;dims=dims)
 end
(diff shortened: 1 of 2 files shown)


ekinakyurek:
@ilkerkesen  I had a different implementation for masked-nll over [here](https://github.com/ekinakyurek/KnetLayers.jl/blob/master/src/loss.jl#L44L78). It is a bit hacky but the advantage is it doesn't require another mask array and probably it is faster because it  has no extra indexing operation. The hack works by forcing user to put 0 into the answer array where the answer will be masked. Then, `findindices` returns location for only non-masked indices.

ilkerkesen:
I liked yours more. I'm just questioning that do we ever need positive mask/pad indices in the output array just for ease of use?

ekinakyurek:
- Thanks @ilkerkesen, and I think there is also a way to support your interface on top of my version of [findindices](https://github.com/ekinakyurek/KnetLayers.jl/blob/master/src/loss.jl#L50). By adding something like below function: 
```
function nll(y, a::AbstractArray{<:Integer}, mask;
             dims=1, average=true)
    a[mask]  .= 0
    indices = findindices(y,masked_answer,dims=dims)
    lp = logp(y,dims=dims)[indices]
    average ? -mean(lp) : -sum(lp)
end
```

- I am not sure that I understand your question correctly:
Is the `output array` predicted scores which is inputs to the `nll`?

ilkerkesen:
`a[mask] .= 0` trick is not feasible because it makes `nll` procedure mutable. I think your initial solution is much more simpler and elegant. Though, I agree with you, an explicit interface where masking doesn't depend on gold output values would be a nice feature to have since it is convenient for evaluation phase.

ekinakyurek:
What do you suggest then? Should I make another PR?

This should work:
 ```JULIA
function nll(y, a::AbstractArray{<:Integer}, mask; dims=1, average=true) = 
  # where mask is boolean or integer array where 0s or falses are masked
    return nll(y,a .* mask;dims=dims, average=average)
end
```

denizyuret:
The least intrusive (backward compatible) change is to implement Ekin's suggestion of skipping zeros in answer key, implemented in: https://github.com/denizyuret/Knet.jl/commit/2f33926ad6aaeac85df18420778cb0b4eb2df28a

This does not break any existing code because zero is currently not a valid answer.

Users will most likely want to use a positive number for padding rather than zero, otherwise the decoder input will be invalid in a s2s model for example. Before calling nll/accuracy the convention is to prepare an answer key where padding is done via zeros. I believe this is the method used in Keras.

Eventually it would be best to implement complete MaskedSequence, PaddedSequence, PackedSequence types like in pytorch. The current solution should be seen as a low level solution. Check out the commit and close PR if satisfactory.

P.S. I had to change the behavior of average=false to return a pair (total,count) rather than just total, as this was needed everywhere. Please check and fix if average=false is used in your code.

PR #488, opened 2019-09-19
new memory manager, more KnetArray functions

denizyuret:
GPU memory improvements
-----------------------
* cuallocator()::Bool can be used to switch between Knet and CuArrays allocators. CuArrays 2x faster, made default.
* New knetgcnode() more aggressively frees GPU pointers during backward pass, decreasing memory footprint 2x.
* minimize returns plain loss again, returning Result prevented gc and increased memory footprint.

New KnetArray functions
-----------------------
* Generalize bmm to handle more than 3 dims.
* KnetArray support for argmax,argmin,findmax,findmin.
* KnetArray support for std,stdm,var,varm.
* Support for both lgamma and loggamma as some people cannot upgrade to SpecialFunctions 0.8.


src/cuarray.jl
```diff
@@ -88,3 +89,33 @@ end
...
+end
+
+# argmax, argmin etc. Fixes https://github.com/denizyuret/Knet.jl/issues/368.
+# Two options: argmax(Array(KnetArray)) vs argmax(CuArray)
+# Experiments: 10x10, 100x100, 1000x1000 with no dims, dims=1, dims=2
+# With no dims, CuArrays is better for 100x100, 1000x1000.
+# With all others, KnetArray is better.
+
+import Base: argmax, argmin, findmax, findmin
+argmaxarray(x,d)=((d===:) && length(x) > 4096 ? CuArray(x) : Array(x))
```
CarloLucibello: Given issue https://github.com/JuliaGPU/CuArrays.jl/issues/304, I think it is better to just use `Array(x)` for all cases

denizyuret: Thanks, good catch. I will switch to Arrays until the issue is fixed.

PR #493, opened 2019-09-23
Attempting to fix build on non-CUDA systems with has_cuda()

ianshmean:


ianshmean:
Ok, I've changed build to throw an error if no NVCC and `find_host_compiler()` fails. That works.

Next to see if the windows, arm and FreeBSD CI images just need a compiler installed!

ianshmean:
How things stand
- It looks like the install process for `gcc` on FreeBSD takes far too long for CI (timed out at 60 mins)
- I don't know how to get Visual Studio set up on appveyor each time

But:
- installation of `gcc` on ARM is working and package tests are passing
- Linux & MacOS continue to work

So I suggest that Cirrus (freeBSD) and Appveyor (Windows) CI is disabled until either CPU-only Knet doesn't depend on a  system compiler, or GPU-enabled CI is available for those platforms.

denizyuret:
Please check out https://github.com/denizyuret/Knet.jl/pull/494

ianshmean:
Superseded by the move to pure-Julia `conv` in the merged #494

PR #494, opened 2019-09-24
Switch to NNlib for CPU conv/pool in pure Julia.

denizyuret:
@ianshmean this should hopefully fix most of the installation problems.


ianshmean:
This is great! 

By the way, the arm image tags in `.drone.yml` needs to be changed to ```image: julia:1.3.0-rc2``` to get past the `var""` error

ianshmean:
Nice 👍 I should've also said that the last 2 drone jobs can be deleted as there isn't an ARM nightly image available.

ianshmean:
Also, as you expected/warned, I just tested `YOLO.jl` with Knet master on a CPU-only machine and saw a slow-down. Speed reduced from ~8 FPS to ~3.8 FPS.

denizyuret:
The couple of conv ops I tried (e.g. with 128x128x3x100 and 5x5x3x50 etc)
did not slow down with NNlib. But I did not do extensive testing. If we can
profile what ops seem slower compared to the C++ version we can bug NNlib
for improvements.

Issue #515, opened 2019-11-06
scaling xavier intializer comparared to other frameworks 


Alexander-Barth:
Firstly, I would like to thank you for all your great work in Knet!

I am trying to debug a convolutional network with Knet which does not perform as well as with Keras/Tensorflow and I noticed that the xavier initialiser does not use the same scaling as the gloriot uniform initializer from Keras (which should be the same as far as I can tell).

In the [original paper](jmlr.org/proceedings/papers/v9/glorot10a/glorot10a.pdf), equation (16), the distribution is defined as:
W ~ U[-scale, scale]
where scale is sqrt(6/(n_j+n_{j+1})} and where n_j and +n_{j+1} is the number of input and output units of a layer.

In Keras/Tensorflow for a 3x3 convolutional layer with 1 input filter and 16 output filters, I get random values of the range -0.1971  and 0.1975 since `sqrt(6/(3*3*1 + 3*3*16)) = 0.1980`.

While for Knet:  `extrema(Knet.xavier(3,3,1,16)) = (-0.2823 0.2786)` since `sqrt(2/(3*3 + 16)) = 0.282842`. Note the 2 instead of the 6 in the numerator and the 16 instead of the 3*3*16.

The documentation links to [caffee](http://caffe.berkeleyvision.org/doxygen/filler_8hpp_source.html) and I belief that class `XavierFiller` does the same as in keras:

```
     int fan_in = blob->count() / blob->num();
     int fan_out = blob->count() / blob->channels();
     Dtype n = fan_in;  // default to fan_in
     if (this->filler_param_.variance_norm() ==
         FillerParameter_VarianceNorm_AVERAGE) {
       n = (fan_in + fan_out) / Dtype(2);
     } else if (this->filler_param_.variance_norm() ==
         FillerParameter_VarianceNorm_FAN_OUT) {
       n = fan_out;
     }
     Dtype scale = sqrt(Dtype(3) / n);
     caffe_rng_uniform<Dtype>(blob->count(), -scale, scale,
        blob->mutable_cpu_data());
```

I tried the following patch in Knet to get the range of weight:

```
diff --git a/src/distributions.jl b/src/distributions.jl
index 90ee383..d35357b 100644
--- a/src/distributions.jl
+++ b/src/distributions.jl
@@ -34,10 +34,13 @@ function xavier(a...)
         fanout = size(w,1)
         fanin = size(w,2)
     else
-        fanout = size(w, ndims(w))
[...]

PR #520, opened 2019-11-19
 xavier intializer using the same convention as TensorFlow and PyTorch

Alexander-Barth:
The PR address issues #515 . Let me know what you think.


denizyuret:
I added some extra stuff, please checkout https://github.com/denizyuret/Knet.jl/pull/522

PR #525, opened 2019-11-29
A numerical example for Knet.nll for the documentation

Alexander-Barth:
I am wondering if the documentation of the function `Knet.nll` would be clearer a numerical example. If the example does not fit, feel free to close the PR without merging.


denizyuret:
Thank you.

PR #526, opened 2019-12-06
fixed primitive for bmm, added more gradient tests for bmm

ekinakyurek:
We missed a one character but critical error in the back function of `bmm` because we didn't test `@diff bmm` for all possible scenarios. Now, I fixed the problem and added required tests.


denizyuret:
Thank you.

PR #541, opened 2020-02-23
respect to CUDA_VISIBLE_DEVICES

ekinakyurek:
In the current master, setting `CUDA_VISIBLE_DEVICES` env. variable does not limit the gpus that Knet can see. For example, if we set `CUDA_VISIBLE_DEVICES=2,3,4,5` in a 8-GPU machine, we got following error.
```julia
┌ Warning: Knet cannot use the GPU: ErrorException("cudaDeviceGetPCIBusId: 101: invalid device ordinal")
└ @ Knet ~/.julia/dev/Knet/src/Knet.jl:167
```

The reason is that the `nvml` library can see all the gpus in the system, where as `cudart` library can only control the gpus with respect to the CUDA_VISIBLE_DEVICES variable. So, I changed the gpu initialization code such that `gpuCount()` always uses `@cudart1` to get the number of gpu devices. 

Now, everything works smoothly,~~except that Knet leaves memory footprint in multiple gpus while it is using a single gpu.~~

~~@denizyuret, you mentioned (in below) that `cudart` leaves the memory footprint, but I observed different result such that `nvml` leaves it. From my experiments, I believe that this function doesn't take any memory~~

https://github.com/denizyuret/Knet.jl/blob/5f0f5461825dbd31b4b76125b4b4d86d107c1d78/src/gpu.jl#L178-L179

~~but this call claims a memory (50MB) in each query~~
https://github.com/denizyuret/Knet.jl/blob/5f0f5461825dbd31b4b76125b4b4d86d107c1d78/src/gpu.jl#L122

~~So, if we solve this issue, the PR is ready to go.~~


ekinakyurek:
I think I fixed the memory footprint issue by adding a device reset before picking the initial gpu. 

FYI: If you also set `export CUDA_DEVICE_ORDER=PCI_BUS_ID`, the cudart ordering becomes same as `nvidia-smi` which is convenient.

denizyuret:
> I think I fixed the memory footprint issue by adding a device reset before picking the initial gpu.

What if the user wants to work with multiple GPUs and keeps switching back and forth between them, is the reset going to mess that up?

> 
> FYI: If you also set `export CUDA_DEVICE_ORDER=PCI_BUS_ID`, the cudart ordering becomes same as `nvidia-smi` which is convenient.

Can we set this via ENV in the __init__() function?

ekinakyurek:
No, It is not. It just does this only at the initialization. I tested.

Yes you can set that as your ENV variable anytime you want. You can decide.

-Ekin
________________________________

denizyuret:
@ekinakyurek When you limit `CUDA_VISIBLE_DEVICES=2,3,4,5`, does cudart see them as 0,1,2,3? (Otherwise the for loops won't work). How about nvml? How does the input/output of nvmlid effected? What is the effect of `export CUDA_DEVICE_ORDER=PCI_BUS_ID` on the different types of ids?

denizyuret:
Ran some tests to answer these questions:
> @ekinakyurek When you limit `CUDA_VISIBLE_DEVICES=2,3,4,5`, does cudart see them as 0,1,2,3? (Otherwise the for loops won't work). 

YES.

> How about nvml? 

nvml still sees them as 2,3,4,5.

> How does the input/output of nvmlid effected? 

nvmlid.([0,1,2,3]) gives [2,3,4,5].

> What is the effect of `export CUDA_DEVICE_ORDER=PCI_BUS_ID` on the different types of ids?

not sure about this one.

PR #549, opened 2020-04-12
GPU discoverability with CUDAapi v4.0

iuliancioarca:
Library versions can be detected automatically by CUDAapi, except for CUDNN which is hardcoded to v7.

### DIFF
--- a/src/gpu.jl
+++ b/src/gpu.jl
@@ -22,12 +22,12 @@ macro cudacall(lib,fun,returntype,argtypes,argvalues,errmsg=true,notfound=:(erro
     lib = string(lib); fun = string(fun)
     if isa(argtypes,Expr); argtypes = argtypes.args; end
     if isa(argvalues,Expr); argvalues = argvalues.args; end
-    if lib=="cudnn" # cudnn version not read automatically. hardcoded to v7
+    if lib == "cudnn" # cudnn version not read automatically. hardcoded to v7
         path = find_cuda_library(lib,tk,[v"7"])
-    elseif lib=="knet8"
+    elseif lib == "knet8"
 	path = libknet8 
     else
-	path = (lib=="knet8" ? libknet8 : find_cuda_library(lib,tk,[ver]))
+	path = find_cuda_library(lib,tk,[ver])
     end
     if path==nothing || path==""; return notfound; end
     fx = Expr(:call, :ccall, Expr(:tuple,fun,path), returntype, Expr(:tuple,argtypes...), argvalues...)
@@ -359,7 +359,7 @@ const cublaserrors = Dict(
 )
 
 function getErrorString(lib,fun,ret)
-    if lib=="cudnn" # cudnn version not read automatically. hardcoded to v7
+    if lib == "cudnn" # cudnn version not read automatically. hardcoded to v7
 	path = find_cuda_library(lib,tk,[v"7"])
     else
     	path = find_cuda_library(lib,tk,[ver])
--- a/src/gpu.jl
+++ b/src/gpu.jl
@@ -1,6 +1,8 @@
 using CUDAapi, TimerOutputs, Libdl
 const libknet8 = Libdl.find_library(["libknet8"], [joinpath(dirname(@__DIR__),"deps")])
 const tk = find_toolkit()
+const tkver = isempty(tk) ? v"0" : parse_toolkit_version(find_cuda_binary("ptxas", tk))# v"0" because "nothing" will crash precompilation
+const cudnnver = v"7" # cudnn version not read automatically. hardcoded to v7
 const to = TimerOutput()
 const Cptr = Ptr{Cvoid}
 function getErrorString end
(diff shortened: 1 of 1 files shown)


iuliancioarca:
I get the serialize error even when testing locally without GPU.
I also don't know how to fix Travis and CI build fails (I suspect they are not configured to use a GPU, since the errors seem to point to that...)

denizyuret:
I will look into this today. Thanks for all the great work!

iuliancioarca:
Actually the first commit had it as a const. I moved it inside the functions because at the beginning I thought that would solve the CI GPU error (`find_cuda_binary("ptxas", tk)` returned an empty string which crashed `parse_toolkit_version`). Unfortunately this only moved the problem inside the functions and led me to think there is no GPU on Travis or Appveyor (I'm not very experienced with this).
I can revert to it, CPU tests are OK:
```
Test Summary: | Pass  Total
Knet          | 2731   2731
    Testing Knet tests passed 
```

denizyuret:
I think having a `tkver` constant makes more sense then. We can defensively define it as:
`const tkver = tk === nothing ? nothing : parse...` to avoid the error you mention.
The only CI with a gpu is gitlab. The others are cpu only machines.
I don't think the intermittent serialization and rnn errors are due to this PR, I am investigating.

iuliancioarca:
Agreed. I will update the code. 
Regarding the RNN error, I got this:
```
Got exception outside of a @test
  curandGenerateUniform: 102: Memory allocation failed
  Stacktrace:
   [1] error(::String) at .\error.jl:33
   [2] curandCreateGenerator(::Int64) at C:\Julia-1.4.0\.julia\packages\Knet\bTNMd\src\gpu.jl:42
   [3] curandGenerator(::Int64; seed::Int64) at C:\Julia-1.4.0\.julia\packages\Knet\bTNMd\src\gpu.jl:238
   [4] seed!(::Int64) at C:\Julia-1.4.0\.julia\packages\Knet\bTNMd\src\random.jl:31
   [5] top-level scope at C:\Julia-1.4.0\.julia\packages\Knet\bTNMd\test\rnn.jl:16
   [6] top-level scope at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.4\Test\src\Test.jl:1113
   [7] top-level scope at C:\Julia-1.4.0\.julia\packages\Knet\bTNMd\test\rnn.jl:7
   [8] include(::String) at .\client.jl:439
   [9] macro expansion at .\util.jl:175 [inlined]
   [10] macro expansion at C:\Julia-1.4.0\.julia\packages\Knet\bTNMd\test\runtests.jl:22 [inlined]
   [11] macro expansion at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.4\Test\src\Test.jl:1113 [inlined]
   [12] top-level scope at C:\Julia-1.4.0\.julia\packages\Knet\bTNMd\test\runtests.jl:22
   [13] include(::String) at .\client.jl:439
   [14] top-level scope at none:6
   [15] eval(::Module, ::Any) at .\boot.jl:331
   [16] exec_options(::Base.JLOptions) at .\client.jl:264
   [17] _start() at .\client.jl:484
```
It chewed up all the 12GB of VRAM on the gpu.

denizyuret:
The remaining problems do not seem to be related to GPU discovery, I am merging this and working on the other problems on a separate branch.

denizyuret:
@iuliancioarca I am trying to test on windows (which I have little experience with), but my nvcc is complaining about not finding the vs compiler:
```
nvcc fatal   : Cannot find compiler 'cl.exe' in PATH
```
Have you had the same problem? I can add this to the path manually but I feel there has to be a better solution...

iuliancioarca:
Hmm...it's been a while since I installed the Build tools. I will give it a try tomorrow: uninstall, remove the path and fresh install the Build tools. It should set it automatically although I'm not 100% sure.

dave7895:
I added it manually. Pretty sure it originally was on the path by itself but I managed to get it out of there somehow so had to get it back manually.

denizyuret:
Can you tell me what exactly do I need to set in the windows environment so I can test and add to the documentation? I tried running vsvars64.bat but then nvcc gives me a bunch of errors.

iuliancioarca:
You can manually set the environment variable path to cl.exe with the next steps:
Hit Windows button and type env
Select Edit system environment variables.
Select the Environment Variables button on the bottom.
Double click on Path variable in the window on top (this sets env. var for your user only).
Now you can add a New env. variable with the path (in my case is: C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\Tools\MSVC\14.21.27702\bin\Hostx64\x64\)

I'm sure there is a shorter way with some powershell commands or bat files but I never bothered with that.
image

RocketRoss:
@denizyuret I have just gone through this, so if you'd like me to document my steps (or just share how to add cl.exe to the path), just ask!

denizyuret:
That would be great. See if the latest docs are accurate for windows and
please suggest changes if not.

dave7895:
to set environment variables that you don't need permanently I personally use 
`set path="%path%;path/to/compiler"` for permanent I use setx, but then it is not available in the current session. not sure if the temporary is enough...

Sh4pe:
Version 1.3.5 does not yet include these patches. They make Knet.jl usable on Windows for me. I'd really love to see a new release that contains this PR.

Thanks all for the great work!

denizyuret:
1.3.6 is now released.

PR #560, opened 2020-05-11
Further notes on GPU tools, especially for windows

RocketRoss:
Simply added an explicit Pkg.test() of all CUDA packages, that windows requires cl.exe from Visual Studio to be in path, and that one can rebuild Knet if gpu tools are installed after knet.


denizyuret:
Thank you.

